Issue #42, opened 2022-09-02
Tests for backward compatibility


srib:
We need tests to ensure backward compatibility primarily for MDIOReader and MDIOWriter.

PR #43, opened 2022-09-02
FSStore dimension separator bug workaround

sanath-2024:
# Patch to Avoid FSStore Bug

## The Bug

Due to "overly aggressive canonicalization" in Zarr's FSStore implementation, all periods ('.') in a logical storage path are replaced with slashes ('/') when the dimension separator is a slash. One major consequence of this bug is that the `.zmetadata` file is implicitly renamed to `/zmetadata` while writing and searching. This has no observable effects on the local filesystem on Linux, since the filesystem concatenates double slashes into a single slash. However, when the object is stored on a cloud object store (e.g. GCS), the double slash is not ignored. Additionally, even on the local filesystem, the concatenation of double slashes is technically an implementation-defined behavior and is therefore not safe. So, it makes sense to avoid the explicit or implicit use of double slashes.

## The Fix

The simplest way to avoid this bug is to avoid using '.' or '/' characters in the path to the metadata file. This change will not affect the ability to read existing MDIO files on the local FS, since the double slash was eliminated by the filesystem anyway. However, it will allow existing MDIO files to be properly read from object stores as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,19 +3,19 @@ updates:
   - package-ecosystem: github-actions
     directory: "/"
     schedule:
-      interval: daily
+      interval: monthly
   - package-ecosystem: pip
     directory: "/.github/workflows"
     schedule:
-      interval: daily
+      interval: monthly
   - package-ecosystem: pip
     directory: "/docs"
     schedule:
-      interval: daily
+      interval: monthly
   - package-ecosystem: pip
     directory: "/"
     schedule:
-      interval: daily
+      interval: monthly
     versioning-strategy: lockfile-only
     allow:
       - dependency-type: "all"


tasansal:
Thanks @sanath-2024

This should fix the problem I think but I need to run some manual tests since we don't have mock object store tests yet. 

There is also another potential solution that may be worth looking into, and would be cleaner if it works:

1. Leave FSStore as is (no separator parameterization)
2. Only apply the slash separator to the arrays

This way we don't break the self describing nature of Zarr. 

I've also mentioned this issue here:

https://github.com/zarr-developers/zarr-python/issues/720#issuecomment-1215302456

sanath-2024:
I think your solution is cleaner. Note that if the FSStore doesn't have a dimension separator it will default back to ".", making the filename ".zmetadata" - on the one hand this is cleaner and it is definitely what's supposed to happen; on the other hand it breaks the existing files that are called "zmetadata" or "/zmetadata".

tasansal:
Created this issue with Zarr:

https://github.com/zarr-developers/zarr-python/issues/1121

In the meantime, @sanath-2024; this is what needs to be done:

1. Revert your changes
2. Remove every usage of `dimension_separator`
3. Use `dimension_separator` ONLY where we create arrays

These are the usages I could find, please take a deeper look.

Let me know if you don't have time to work on this, I can make changes and commit here too.

https://github.com/TGSAI/mdio-python/blob/509fdcb94c1c09466667656d1bc027949dee75d9/tests/unit/conftest.py#L104-L109

https://github.com/TGSAI/mdio-python/blob/811c10581c1f251ad6b9d15ec9ee8bf54145cd05/src/mdio/segy/blocked_io.py#L81-L87

https://github.com/TGSAI/mdio-python/blob/811c10581c1f251ad6b9d15ec9ee8bf54145cd05/src/mdio/segy/blocked_io.py#L110-L116

https://github.com/TGSAI/mdio-python/blob/e23ec72d77b7b7dc6d09ac2f7519951fd74c9439/src/mdio/converters/segy.py#L209-L214

tasansal:
closes #42 

This PR will eliminate and backwards compatibility issues until Zarr v3 or we make major changes

sanath-2024:
I can make those changes, and I'll make sure to run `pre-commit` next time. I will keep you updated about my progress.

PR #46, opened 2022-09-04
Improved ZFP compression quality using `FIXED-ACCURACY` mode

tasansal:
We used to use `FIXED-RATE` compression to achieve predictable compression ratios. However, this does introduce some large errors.

After benchmarking, we see that `FIXED-ACCURACY` with a tolerance achieves similar compression ratios with significantly fewer compression artifacts. 

This PR changes the compression method and set a reasonable default.

The default value of `0.01` will guarantee that compressed data will have reconstruction errors less than that value. Typically this compresses seismic data to 20-25% of its size.

The results with a tolerance of `0.1` or higher like `1` are also acceptable for visualization applications and provide about 90% or more reduction in size.

Note that this number must be reduced if the dataset has very small values/variations. The data will become less compressible if we store very small numbers (like `1e-5` levels).

References:
https://zfp.readthedocs.io/en/release1.0.0/introduction.html
https://www.blosc.org/posts/support-lossy-zfp/

- [X] Tests passes
- [X] Documentation is updated
- [X] Linter passes

### DIFF
--- a/src/mdio/commands/segy.py
+++ b/src/mdio/commands/segy.py
@@ -123,12 +123,12 @@ cli = click.Group(name="segy", help=SEGY_HELP)
     show_default=True,
 )
 @click.option(
-    "-ratio",
-    "--compression-ratio",
+    "-tolerance",
+    "--compression-tolerance",
     required=False,
-    default=4,
-    help="Lossy compression ratio.",
-    type=click.INT,
+    default=0.01,
+    help="Lossy compression tolerance in ZFP.",
+    type=click.FLOAT,
     show_default=True,
 )
 @click.option(
@@ -156,7 +156,7 @@ def segy_import(
     chunk_size,
     endian,
     lossless,
-    compression_ratio,
+    compression_tolerance,
     storage_options,
     overwrite,
 ):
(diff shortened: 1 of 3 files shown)


tasansal:
The below compresses a Volve dataset stack (ST1001) with _MDIO_ using lossy compression. 
This is the default compression only by using `lossless=False` in Python API or `--lossless False` in CLI: 
Note that the fourth panel is 1,000 times gained difference.

Size Comparison
| Version              | Size    | Ratio  |  PSNR¹  |     SSIM²    |     MSE³     |  NRMSE⁴  | Description                                                                                     |
|----------------------|---------|--------|:------:|:-----------:|:-----------:|:-------:|-------------------------------------------------------------------------------------------------|
| SEG-Y                | 1300 MB | 1:1    | ∞      | 1.0         | 0.0         | 0.0     | Original SEG-Y (Ground Truth)                                                                   |
| MDIO (Lossless)      | 998 MB  | 0.77:1 | ∞      | 1.0         | 0.0         | 0.0     | Lossless                                                                                        |
| MDIO (Lossy Default) | 264 MB  | 4:1 | 106.69 | 0.999999978 | 0.000000092 | 0.00033 | Perceptually lossless, can still be used for numerical calculations (depending on requirements) |
| MDIO (Lossy tol=1)   | 53 MB   | 25:1 | 66.28  | 0.999721336 | 0.001010011 | 0.03467 | More lossy, mainly for visualization.                                                           |

Metrics are calculated from the inline in the images below. 
¹ Peak signal-to-noise ratio. Higher is better. (max ∞) (normalized by whole dataset range) 
² Structural similarity index. Higher is better. (max 1) 
³ Mean square error 
⁴ Normalized root mean square error (normalized by whole dataset range) 

PSNR values above 60 are considered acceptable in the lossy image compression community.

# Default Lossy (tolerance = 0.01)
image

# More Lossy (tolerance = 1)
image

PR #52, opened 2022-09-06
Use dimension separator in arrays instead of `FSStore`

sanath-2024:
Move all usage of `dimension_separator` from Zarr stores to individual Zarr arrays, in order to avoid `.zmetadata` dimension separator bug

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@
 /dist/
 /docs/_build/
 /src/*.egg-info/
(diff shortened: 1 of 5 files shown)


srib:
Closes #42

tasansal (changes requested):

tasansal (approved):

tasansal:
thanks @sanath-2024

PR #60, opened 2022-09-08
update hyperlink to work with readthedocs

tasansal:


tasansal:
github integration isn't working

PR #62, opened 2022-09-20
Make `storage_options` consistent with fsspec to be able to pass local cache directory

sanath-2024:
By default, the `simplecache` protocol in `fsspec` creates a new temporary directory to serve as the local cache for a remote filesystem. However, if we want to use an existing local cache, we need the ability to specify an existing directory as the local cache.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -30,21 +30,21 @@ fsspec = ">=2021.10.1"
 
 [[package]]
 name = "aiobotocore"
-version = "2.3.4"
+version = "2.4.0"
 description = "Async client for aws services using botocore and aiohttp"
 category = "main"
 optional = true
-python-versions = ">=3.6"
+python-versions = ">=3.7"
 
 [package.dependencies]
 aiohttp = ">=3.3.1"
 aioitertools = ">=0.5.1"
-botocore = ">=1.24.21,<1.24.22"
+botocore = ">=1.27.59,<1.27.60"
 wrapt = ">=1.10.10"
 
 [package.extras]
-awscli = ["awscli (>=1.22.76,<1.22.77)"]
-boto3 = ["boto3 (>=1.21.21,<1.21.22)"]
+awscli = ["awscli (>=1.25.60,<1.25.61)"]
+boto3 = ["boto3 (>=1.24.59,<1.24.60)"]
 
 [[package]]
 name = "aiohttp"
@@ -64,11 +64,11 @@ multidict = ">=4.5,<7.0"
 yarl = ">=1.0,<2.0"
 
 [package.extras]
-speedups = ["aiodns", "brotli", "cchardet"]
+speedups = ["Brotli", "aiodns", "cchardet"]
 
 [[package]]
 name = "aioitertools"
-version = "0.10.0"
+version = "0.11.0"
 description = "itertools and builtins for AsyncIO and mixed iterables"
 category = "main"
 optional = true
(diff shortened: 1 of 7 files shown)


sanath-2024:
For 1, the `simplecache` test wouldn't run until I installed `lockfile`. However, strangely, when I uninstalled `lockfile` the test still ran fine! I will revert the dependencies. I think Poetry was complaining because I had an outdated installation of `lockfile`, since that's the only relevant code I can find that uses `lockfile`.

For 2, I can see a couple of options to pass the local cache directory within the `storage_options` dict itself:

Existing behavior:

```py
# no disk cache: leave options as is
process_url(url="s3://some/path", storage_options={"some key": "some value"}, disk_cache=False, ...)

storage_options == {"some key": "some value"}
```

```py
# disk cache: modify storage options
process_url(url="s3://some/path", storage_options={"some key": "some value"}, disk_cache=True, ...)

storage_options == {"s3": {"some key": "some value"}}
```

Option 1: never modify the input `storage_options`

```py
# no disk cache: leave options as is
process_url(url="s3://some/path", storage_options={"some key": "some value"}, disk_cache=False, ...)

storage_options == {"some key": "some value"}
```

```py
# disk cache: leave options as is (make the user specify their S3 storage options under the "s3" key)
process_url(url="s3://some/path", storage_options={"s3": {"some key": "some value"}}, disk_cache=True, ...)

storage_options == {"s3": {"some key": "some value"}}
```

```py
# disk cache with custom storage directory: leave options as is (make the user specify which options are going to each FS)
process_url(url="s3://some/path", storage_options={"s3": {"some key": "some value"}, "simplecache": {"cache_storage": "/local/path"}}, disk_cache=True, ...)

storage_options == {"s3": {"some key": "some value"}, "simplecache": {"cache_storage": "/local/path"}}
```

Option 2: move the `cache_storage` option to where it needs to go

```py
# no disk cache: leave options as is
process_url(url="s3://some/path", storage_options={"some key": "some value"}, disk_cache=False, ...)

storage_options == {"some key": "some value"}
```

```py
# disk cache: modify storage options
process_url(url="s3://some/path", storage_options={"some key": "some value"}, disk_cache=True, ...)

storage_options == {"s3": {"some key": "some value"}}
```

```py
# disk cache with custom storage directory: assume that only the `cache_storage` option is going to `simplecache`, put all other options in `s3`, `gcs`, etc.
process_url(url="s3://some/path", storage_options={"some key": "some value", "cache_storage": "/local/path"}, disk_cache=True, ...)

storage_options == {"s3": {"some key": "some value"}, "simplecache": {"cache_storage": "/local/path"}}
```

Option 1 would be more flexible, since users could specify which options go to which FS. However, option 2 would be more compatible with existing solutions and would be easier to use (although its interface would be significantly different from the `fsspec` interface).

Which one would you recommend? Or were you thinking of something else?

tasansal:
I am leaning towards option 1 to be directly compatible with `fsspec`. 

Option 2 is nice too but it is less explicit. 

We need to write docs with examples for either. 

I am voting option 1, let's ask what @srib thinks.

srib:
I think option 1 is easy to understand, explicit and compatible with
`fsspec`. I would prefer that over option 2.

In any case, we need to document the use case well because of too many
key-value pairs in the storage options!

tasansal:
Thanks @sanath-2024! Code changes look good to me. Can we move the comments into the doctoring and add some in the examples section? 

See here for docstring example.

I will merge this after the docstring update but needs to be tested with combining a cloud provider options and simple cache before next release.

tasansal:
@sanath-2024 I suggested adding the doctest style strings, so it gets rendered as code in the documentation. Also, added slightly more detail to the docstring you wrote. LGTM besides that!

tasansal (changes requested):

tasansal:
Oops broke the linter, I'll fix it and merge.

tasansal (approved):

PR #109, opened 2022-10-11
Refactor SEG-Y export to optimize memory usage.

### LINKED ISSUE #34: Reduce memory consumption during SEG-Y export
The distributed workers flatten the chunks along the first dimension to write to SEG-Y.

Huge files >2TB use a lot of memory during export.

The output sharding strategy needs to be optimized:

https://github.com/TGSAI/mdio-python/blob/03b9e4f7e3ef6a149c4b6c0980f544fb1b72afef/src/mdio/converters/mdio.py#L177-L241

and

https://github.com/TGSAI/mdio-python/blob/03b9e4f7e3ef6a149c4b6c0980f544fb1b72afef/src/mdio/segy/creation.py#L111

tasansal:
closes #34 

ref: [Dask Discourse](https://dask.discourse.group/t/serializing-3d-dataset-with-chunks-to-sequential-binary-on-disk-without-breaking-order/1172)

Proposed graph for 3D below. 2D or 4D is similar, with less or more hierarchy. 

There are a few minor changes not related to export.

image


tasansal:
This has significant improvements to memory usage:

https://github.com/dask/distributed/discussions/7128

```python
import dask
import distributed

with dask.config.set({"distributed.scheduler.worker-saturation": "1.0"}):
    client = distributed.Client(...)
```

tasansal:
@srib can you take a look when you have a chance, the entire export logic changed here, would be good to have a second set of eyes.

PR #248, opened 2023-07-26
Support for Seismic Data Ingestion Without Dimension-Specific Indexing

### LINKED ISSUE #241: Support for segy ingestion of non-regularized data
Some types of segy data do not have a discrete integer bin on input.  E.g. 2D CDP data might be regular in CDP and time but the offset dimension maybe represented by a floating point offset.  

To handle cases like this the proposal is to have an ingestion header "index" which will just use a counter starting at 1and append traces to the "index" header.

### LINKED ISSUE #242: parse_trace_headers bug
parse trace headers should return an nd array.  It currently returns a dictionary as implemented in the legacy solution due to an error introduced in #213 and for floating point header support (#212)

markspec:
Resolves #241 also fixes typing in a number of places that might also resolve #242.  A new "AutoIndex" grid override has been implemented that can be used in conjunction with the "trace" index name to automatically generate a trace index and cn be used to ingest non-regularized data.


tasansal:
@markspec we need to change the logic here. When we ingest this way, we are losing self-describing information from the MDIO file. i.e. it is unclear what indices the `trace` key flattened. This information gets lost.

We need to add dimensions without coordinates; i.e. just dimensions. This way they won't mean anything but we should preserve index keys names. Some examples:
```python
("inline", "crossline", "offset", "azimuth")
```
becomes 
```python
("inline", "crossline", "trace")
```
and there is no record of what went into trace. Instead we should still ingest it as:
```python
("inline", "crossline", "offset", "azimuth")
```
However, it should know that `offset` and `azimuth` as not indexed. We should implement a simpler version of non-index and index dimensions concept like `Xarray` has. See [here](https://docs.xarray.dev/en/stable/user-guide/terminology.html).

tasansal:
@markspec after speaking w/ Santh, Cable, and @srib we decided on the following. The `NonBinned` is explicit; and we preserve the names and bytes the user intended to use if it was gridded. This way during data exploration; users know the lineage of the data. We can save "names" and "bytes" into the coordinate's attributes with another key/value pair as "non_binned" = True.

### Non-Binned 3D Seismic Offset Example
```python
segy_to_mdio(
    ...,
    index_bytes=(189, 193),
    index_names=("inline", "crossline"),
    chunksize=(4, 4, 1024),
    grid_overrides={"NonBinned": True, "chunksize": 16, "names": ["offset"], "bytes": [37]}
)
```

### Non-Binned 3D Seismic Offset / Azimuth Example
```python
segy_to_mdio(
    ...,
    index_bytes=(189, 193),
    index_names=("inline", "crossline"),
    chunksize=(4, 4, 1024),
    grid_overrides={"NonBinned": True, "chunksize": 16, "names": ["offset", "azimuth"], "bytes": [37, 205]}
)
```

### Non-Binned 2D Seismic Offset Example
```python
segy_to_mdio(
    ...,
    index_bytes=(21,),
    index_names=("cdp"),
    chunksize=(4, 1024),
    grid_overrides={"NonBinned": True, "chunksize": 16, "names": ["offset"], "bytes": [37]}
)
```

### Binned 2D Seismic Offset with Duplicate Traces Example
```python
segy_to_mdio(
    ...,
    index_bytes=(21, 37),
    index_names=("cdp", "offset"),
    chunksize=(4, 16, 1024),
    grid_overrides={"HasDuplicates": True}  # chunksize would be set to 1; and "trace" dimension will have "duplicate" = True attribute
)
```

There may be some things we are all missing; but I'm sure they'll come out while being implemented.

markspec:
@tasansal just a note on implications.  Currently GridOverrides is setup to modify and return index_headers.  To implement your suggested API change would require a refactor of GridOveride to return and modify index_names and chunksize.

markspec:
@tasansal , get_grid_plan() would need to modify the chunk_size which is very ugly!

tasansal:
made some changes to the examples per our discussion

tasansal:
`NonBinned` and `HasDuplicates` can use the same `AutoIndex` grid override and be specified in `OverrideHandler`; and chunksize defaults to 1.

markspec:
@tasansal , please can you take a look at the safety test.

tasansal:
> @tasansal , please can you take a look at the safety test.

looks like the `poetry lock` fixed the issue. I am working on some minor fixes, will merge the PR today.

tasansal:
@markspec 

Please check this diff (from where you left off to my final version). It's ready to merge unless you see something I messed up. All tests pass.

https://github.com/TGSAI/mdio-python/compare/f6b64f2fa08435ba50f01ae84f07bcda7d10a895..markspec:mdio-python:3e401e2713871e363948fe67cec1e8f3b582c089?diff=split

markspec:
@tasansal , looks good to me.

tests/integration/test_segy_import_export.py
```diff
@@ -12,17 +12,77 @@
...
+        grid = mdio.grid
+
+        assert grid.select_dim(header_names[0]) == Dimension(shots, header_names[0])
+        assert grid.select_dim(header_names[1]) == Dimension(cables, header_names[1])
+        assert grid.select_dim("trace") == Dimension(
+            range(1, np.amax(receivers_per_cable) + 1), "trace"
+        )
+        samples_exp = Dimension(range(0, num_samples, 1), "sample")
+        assert grid.select_dim("sample") == samples_exp
+
```
tasansal, line 76: I just noticed, on any of these, we don't verify if the sample values (amplitudes) and exported SEG-Ys are correct. Can we add them?

Amplitude checks would go to TestReader class in this file; and exports would go to TestExport class

tasansal (approved):

tests/integration/test_segy_import_export.py
tasansal: @markspec let's make a new PR for these

Issue #331, opened 2024-01-04
Improve logging for autochannelwrap


markspec:
Improve logging to let users know the reason for sparse grid and take self-corrective action on whether to ingest with autoindex or autochannelwrap.

This has become an issue for streamer data where ingestion with autochannelwrap usually works.  In the unusual case of unwrapped channels with duplicate traces then segy2mdio ingestion will fail with a sparse index error.  In this case the user would like to see an what the first repeated channel was to aid diagnosis and a suggestion to use autoindex in the case that the job fails.

tasansal:
💯 for improving logging.

I would be against taking self-corrective action by the code. It makes it unpredictable.

tasansal:
merged pr, closing.

PR #332, opened 2024-01-04
Improved logging in autochannelwrap.

markspec:
Update the logging in autochannlewrap.

It now logs explaining when and why it is using type A. It also suggests corrective actions if the user expects it to ingest as type b (unwrapped).

Resolves issue #331.


tasansal:
@markspec I made some code style changes to make it more readable. All tests pass, but to ensure logic didn't break, can you double-check? It's ready to merge besides that.

tasansal (approved):
approved with minor changes

markspec:
@tasansal , looks good!

PR #526, opened 2025-04-04
Python 3.13 with uv

BrianMichell:
Migration from Poetry to uv fixes dependency resolution error for NumPy.
Fixes:
- Incorrect NumPy version resolution
- Undefined behavior due to source-compilation

From #525


tasansal:
closing this and will continue on #496.

PR #527, opened 2025-04-08
Chunked live mask

BrianMichell:
- Adds optional user-defined chunking to the live mask on segy ingestion.

This PR does not update the rechunking function.


tasansal:
@BrianMichell i don't like the idea of adding another parameter. Can we automatically chunk it if buffer size is >2GB ?

E.g. If live mask is >2GB, chunk it into 4 pieces.

This also makes me think, the trace counter may have an issue too after this change. See here:
https://github.com/TGSAI/mdio-python/blob/db741ea4c4cd09c8bd2b54819955347a014dfec3/src/mdio/core/grid.py#L92

This limits max number of traces to uint32 max, e.g. ~4 billion. We should switch that to UINT64 probably, but please check all usages and ensure it doesn't cause memory consumption issues.

BrianMichell:
Automatic chunking implemented.

I also added some logic for automatically promoting the trace counter dtype if we run into that case. I don't like the idea of doubling the memory requirement if it's not necessary all (most) the time.

src/mdio/converters/segy.py
```diff
@@ -494,3 +497,43 @@ def segy_to_mdio(  # noqa: C901
...
+    """Calculate the optimal chunksize for the live mask.
+
+    Args:
+        grid: The grid to calculate the chunksize for.
+
+    Returns:
+        Either a single integer (-1) if no chunking is needed, or a sequence of integers
+        representing the optimal chunk size for each dimension of the grid.
+    """
+    if np.sum(grid.live_mask) < INT32_MAX:
```
tasansal, line 512: Mask can be still >2GB and have only 1 live sample. It won't catch the case. Below should work

```suggestion
    if grid.live_mask.nbytes < INT32_MAX:
```

src/mdio/converters/segy.py
```diff
@@ -494,3 +497,43 @@ def segy_to_mdio(  # noqa: C901
...
+        # Base case where we don't need to chunk the live mask
+        return -1
```
tasansal, line 514: If we return the full shape we can always return a Sequence[int], so makes the `| int` unnecessary

```suggestion
        return grid.live_mask.shape
```

src/mdio/converters/segy.py
```diff
@@ -494,3 +497,43 @@ def segy_to_mdio(  # noqa: C901
...
+
+    # Calculate the optimal chunksize for the live mask
+    total_elements = np.prod(grid.shape[:-1])  # Exclude sample dimension
```
tasansal, line 517:
```suggestion
    total_elements = grid.live_mask.size

File: src/mdio/converters/segy.py [hunk 4]
```diff
@@ -494,3 +497,43 @@ def segy_to_mdio(  # noqa: C901
...
+    num_chunks = np.ceil(total_elements / INT32_MAX).astype(int)
+
+    # Calculate chunk size for each dimension
+    chunks = []
+    remaining_elements = total_elements
+
+    for dim_size in grid.shape[:-1]:  # Exclude sample dimension
```
tasansal, line 524:
```suggestion
    for dim_size in grid.live_mask.shape:  # Exclude sample dimension
```

src/mdio/converters/segy.py
```diff
@@ -494,3 +497,43 @@ def segy_to_mdio(  # noqa: C901
...
+        # Calculate how many chunks we need in this dimension
+        # We want to distribute chunks evenly across dimensions
+        dim_chunks = max(
+            1,
+            int(
+                np.ceil(
+                    dim_size / np.ceil(np.power(num_chunks, 1 / len(grid.shape[:-1])))
+                )
+            ),
+        )
```
tasansal, line 534: this looks a bit confusing, can we rewrite it differently?

src/mdio/core/grid.py
```diff
@@ -88,8 +90,26 @@ def build_map(self, index_headers):
...
+            dtype = "uint64"
+            fill_value = UINT64_MAX
+        else:
+            dtype = "uint32"
+            fill_value = UINT32_MAX
+
+        # We set dead traces to max uint32/uint64 value.
+        # Should be far away from actual trace counts.
+        self.map = zarr.full(self.shape[:-1], dtype=dtype, fill_value=fill_value)
         self.map.vindex[live_dim_indices] = range(len(live_dim_indices[0]))
```
tasansal, line 114: I dont see where `_calculate_live_mask_chunksize` is used?

tasansal (changes requested):
I think it looks good but we can clean it up a little bit.

Also, can you update the PR description and include the pseudo-code for the algorithm? I'm not 100% sure how we are determining chunks. 

Maybe we can reuse Dask's auto chunker? Check segy-export; we use that there a lot.

src/mdio/core/grid.py
BrianMichell: `_calculate_live_mask_chunksize` is only used in `converters/segy.py`. I don't think it's relevant for grid calculations.

tasansal:
After discussion we decided to:

* Set `live_mask` chunk sizes to be max 0.5Gi in memory (smaller on disk ofc). Based on:
  * If users are reading subsets of `live_mask`, they need less memory
  * Nicer chunking behaviour
* Use Dask's rechunker logic with balanced chunk sizes.

Somthing like:

```python
from dask.array.core import normalize_chunks
from dask.array.rechunk import _balance_chunksizes


chunks = normalize_chunks(
    "auto",
    shape,
    dtype="bool",
    limit="0.5Gi",
)
final_chunks = tuple(_balance_chunksizes(chunk)[0] for chunk in chunks)
```

tasansal:
I made some changes please review when you get a chance.

* Simplified the auto chunking functions and consolidated them to `core.utils_write`
* Simplified `grid_map` pre-optimization
* Optimized `grid_map` to process in batches instead of all at once
* Consolidated tests to only 3 functions with `pytest.mark.parametrize`
* Changed one of the test's logic to make sure nbytes is +/-25% of 500MB for the fudge

tasansal (approved):

PR #529, opened 2025-04-10
Support for MDIO Origination from Scratch or Existing Dataset

tasansal:
We can now create MDIO files from scratch!

See the docs page for more details.

See docs build here based on this PR: https://mdio-python--529.org.readthedocs.build/en/529/notebooks/creation.html

Main Changes:
* New `create_empty` function to create MDIO from scratch
* New `create_empty_like` function to clone an existing MDIO file without trace, header, stat, and live_mask data; to be filled later.
* Configuration dataclasses for creating the MDIO file 
* SEG-Y ingestion uses the new create_empty and updates only things necessary
* Fix lots of issues in `MDIOWriter` to be able to populate new MDIO
* Tests updated to use new creation logic instead of manual mock
* Added documentation page to show how its used


src/mdio/api/accessor.py
```diff
@@ -646,7 +649,7 @@ def __init__(
         """Initialize super class with `r+` permission."""
```
BrianMichell, line 649:
```suggestion
        """Initialize super class with `w` permission."""
```

BrianMichell (changes requested):

src/mdio/api/accessor.py
```diff
@@ -365,12 +362,17 @@ def chunks(self, value: tuple[int, ...]) -> None:
     @property
     def stats(self) -> dict:
         """Get global statistics like min/max/rms/std."""
-        return self._stats
+        keys = {"mean", "std", "rms", "min", "max"}
```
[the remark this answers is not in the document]
tasansal: true, i'll put it in as a class attribute

src/mdio/api/accessor.py
tasansal: fixed

BrianMichell (approved):
LGTM.

Will break other pending PRs, but should be minimal effort to rebase and update.

src/mdio/core/factory.py
```diff
@@ -0,0 +1,258 @@
...
+        dtype: The data type of the variable (e.g., 'float32', 'int16').
+        chunks: The chunk size for the variable along each dimension.
+        compressor: The compression algorithm to use (e.g., Blosc or ZFPY).
+        header_dtype: The data type for the variable's header.
+    """
+
+    name: str
+    dtype: str
+    chunks: tuple[int, ...] | None = None
+    compressor: Blosc | ZFPY | None = None
```
markspec, line 73: Would it be cleaner to define a compressor type rather than having a long list here?  This will get ugly really quickly.

src/mdio/core/factory.py
```diff
@@ -0,0 +1,258 @@
...
+            False, an error is raised if the dataset exists. Defaults to False.
+        storage_options: Options for the storage backend, such as credentials or settings for
+            cloud storage (e.g., S3, GCS). Defaults to None.
+        consolidate_meta: If True, consolidates metadata into a single file after creation,
+            improving performance for large metadata. Defaults to True.
+
+    Returns:
+        Group: The root Zarr group representing the newly created MDIO dataset.
+    """
+    storage_options = storage_options or {}
```
markspec, line 125: storage_options: dict[str, Any]  = {}, would enable you to remove the requirement for this line.

src/mdio/core/factory.py
```diff
@@ -0,0 +1,258 @@
...
+            name=f"{variable.name}_trace_headers",
+            shape=config.grid.shape[:-1],  # Same spatial shape as data
+            chunks=variable.chunks[:-1],  # Same spatial chunks as data
+            compressor=Blosc("zstd"),
+            dtype=header_dtype,
+            dimension_separator="/",
+            write_empty_chunks=False,
+        )
+
+    stats = {
```
markspec, line 183: We should not populate with incorrect values.  It would be better to populate will None.

markspec (commented):
Do we also need to add some error handling?

src/mdio/core/factory.py
tasansal: We can't do that. Using a mutable default dictionary in a function definition can lead to unexpected behavior. Checking for `None` and assigning a default is standard practice.

src/mdio/core/factory.py
tasansal: Yes, I'll fix.

src/mdio/core/factory.py
tasansal: Do you mean this for the stats? The stats being 0 is correct because empty array's fill value is 0. Write empty chunks will fill array with fill value once its read.

PR #551, opened 2025-05-19
Update zarr dependency version range

tasansal:
Zarr fixed some bugs that caused us to pin the upper limit. Now there's a new version that fixes things, so that's now our min dependency.


tasansal:
@BrianMichell can you please test this in rigor and let me know if previous issues are gone; I'll merge after.

BrianMichell:
All tests pass and the issues from #533 appear to be resolved with Zarr 3.0.8.

LGTM

PR #553, opened 2025-05-30
V1 schema review

BrianMichell:
Coordinate metadata was required to be a list of units and attributes. This doesn't seem sensible since it only accepts a Scalar data type.

This PR updates a Coordinate to take a UnitsV1 object (singleton or list of) and UserAttributes. It also helps document when to define a Coordinate or a Variable.


BrianMichell:
@tasansal

PR #558, opened 2025-06-03
497 ingestion memory

### LINKED ISSUE #497: Ingestion memory usage
During some internal benchmarking we noticed some interesting behavior with memory usage. @tasansal thinks it could be related to [vindex](https://github.com/TGSAI/mdio-python/blob/927cd3fcd1a1b544278a8d983fe2050878d6bd11/src/mdio/core/grid.py#L80C1-L96C52).

Image

BrianMichell:
Closes #497 

Resolves memory usage issues by lazily computing indices instead of eagerly computing the full `vindex` array.

Before:
image

After:
image

### DIFF
--- a/src/mdio/converters/segy.py
+++ b/src/mdio/converters/segy.py
@@ -354,11 +354,13 @@ def segy_to_mdio(  # noqa: PLR0913, PLR0915, PLR0912
         ...     grid_overrides={"HasDuplicates": True},
         ... )
     """
-    from zarr.core.config import config as zarr_config
     import os
-    num_cpus = int(os.getenv("MDIO__IMPORT__CPU_COUNT", 1))
+
+    from zarr.core.config import config as zarr_config
+
+    num_cpus = int(os.getenv("MDIO__IMPORT__CPU_COUNT", "1"))
     zarr_config.set({"threading.max_workers": num_cpus})
-    
+
     index_names = index_names or [f"dim_{i}" for i in range(len(index_bytes))]
     index_types = index_types or ["int32"] * len(index_bytes)
 


src/mdio/segy/blocked_io.py
```diff
@@ -78,34 +87,26 @@ def to_zarr(segy_file: SegyFile, grid: Grid, data_array: Array, header_array: Ar
...
-        # This executes the lazy work.
         chunk_stats = list(lazy_work)
 
-    # This comes in as n_chunk x 5 columns.
-    # Columns in order: count, sum, sum of squared, min, max.
-    # We can compute global mean, std, rms, min, max.
-    # Transposing because we want each statistic as a row to unpack later.
-    # REF: https://math.stackexchange.com/questions/1547141/aggregating-standard-deviation-to-a-summary-point  # noqa: E501
-    # REF: https://www.mathwords.com/r/root_mean_square.htm
+    # Aggregate statistics
```
tasansal, line 92: comments are gone, good, but re-add the reference links please

src/mdio/segy/blocked_io.py
```diff
@@ -48,21 +57,21 @@ def to_zarr(segy_file: SegyFile, grid: Grid, data_array: Array, header_array: Ar
...
+    # Determine number of workers
+    num_cpus_env = int(os.getenv("MDIO__IMPORT__CPU_COUNT", default_cpus))
+    num_workers = min(num_chunks, num_cpus_env)
     context = mp.get_context("spawn")
-    executor = ProcessPoolExecutor(max_workers=num_workers, mp_context=context)
 
     # Chunksize here is for multiprocessing, not Zarr chunksize.
     pool_chunksize, extra = divmod(num_chunks, num_workers * 4)
-    pool_chunksize += 1 if extra else pool_chunksize
+    pool_chunksize += 1 if extra else 0
```
tasansal, line 67: this is interesting; what was the behaviour before?

Was it adding pool_chunksize if not extra?

Maybe simpler to

```python
if not extra:
    pool_chunksize += 1
```

wdyt

src/mdio/segy/blocked_io.py
```diff
@@ -48,21 +57,21 @@ def to_zarr(segy_file: SegyFile, grid: Grid, data_array: Array, header_array: Ar
...
     num_chunks = len(chunker)
 
-    # For Unix async writes with s3fs/fsspec & multiprocessing, use 'spawn' instead of default
-    # 'fork' to avoid deadlocks on cloud stores. Slower but necessary. Default on Windows.
-    num_cpus = int(os.getenv("MDIO__IMPORT__CPU_COUNT", default_cpus))
-    num_workers = min(num_chunks, num_cpus)
+    # Determine number of workers
+    num_cpus_env = int(os.getenv("MDIO__IMPORT__CPU_COUNT", default_cpus))
+    num_workers = min(num_chunks, num_cpus_env)
     context = mp.get_context("spawn")
```
tasansal, line 63: maybe move this down too, next to ProcessPoolExecutor

src/mdio/core/grid.py
```diff
@@ -106,47 +109,64 @@ def from_zarr(cls, zarr_root: zarr.Group) -> Grid:
         return cls(dims_list)
 
     def build_map(self, index_headers: HeaderArray) -> None:
```
tasansal, line 111: i don't understand the need to change in this function and the next. the map and live_mask are compressed zarr arrays in memory and headers are already processed in batch.

tasansal (changes requested):
i think we over complicated it and I don't understand what component contributes to reducing memory footprint. Some things that were compressed in memory are now full allocation etc. Can you please explain the algorithmic changes and maybe we can revert some that doesn't help with memory footprint

src/mdio/converters/segy.py
```diff
@@ -446,13 +452,58 @@ def segy_to_mdio(  # noqa: PLR0913, PLR0915
...
+    # Clean up live mask related variables
+    del live_mask_array
+    del chunker
+    del block
+    del local_coords
+    del trace_ids
+    del chunk_indices
+    import gc
+    gc.collect()
+
```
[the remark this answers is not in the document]
BrianMichell: Deeper dive into why I did this

Here's a zoomed in view of commit `4274561b4eebab1894509d17d7520d74e5fa87d2` (No eager del or gc)
image

Here's a zoomed in view after adding the eager garbage collection
image

These two plots show the period between `from mdio.core.indexing import ChunkIterator` and `# Clean up live mask related variables` in the current code.

It helped decrease heap allocation by ~200MiB to eagerly delete and garbage collect the objects.

*I failed to drop the cache on the `After` run of the initial PR message. These figures are both with dropped cache.

src/mdio/core/grid.py
BrianMichell: This was actually the majority of the reason behind the change. 

The `vindex` used to eagerly generate the map appeared to be responsible for the sharp spike around 11:05 in the initial figure. The intent of these changes was to effectively defer those computations to the distributed tasks.

```python
        for start in range(0, total_live_traces, batch_size):
            end = min(start + batch_size, total_live_traces)
            live_dim_indices = []

            # Compute indices for the batch
            for dim in self.dims[:-1]:
                dim_hdr = index_headers[dim.name][start:end]
                indices = np.searchsorted(dim, dim_hdr).astype(np.uint32)
                live_dim_indices.append(indices)
            live_dim_indices = tuple(live_dim_indices)

            # Assign trace indices
            trace_indices = np.arange(start, end, dtype=np.uint64)
            self.map.vindex[live_dim_indices] = trace_indices  # <--- This vindexing assignment 
            self.live_mask.vindex[live_dim_indices] = True
```

src/mdio/converters/segy.py
```diff
@@ -446,13 +452,58 @@ def segy_to_mdio(  # noqa: PLR0913, PLR0915
...
+            # Only convert dtype if necessary for indexing (numpy requires int for indexing)
+            if local_idx.dtype != np.intp:
+                local_idx = local_idx.astype(np.intp)
+            local_coords.append(local_idx)
+
+        # Mark live cells in the temporary block
+        block[tuple(local_coords)] = True
+
+        # Write the entire block to Zarr at once
+        live_mask_array.set_basic_selection(selection=chunk_indices, value=block)
```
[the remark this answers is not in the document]
BrianMichell: Correction: I confused myself between the internal Grid chunking and the `utils_write.py` chunking size. This does leverage the persistent live_mask chunk size appropriately. `block` should be 512MiB as defined by `MAX_SIZE_LIVE_MASK`.

src/mdio/segy/blocked_io.py
BrianMichell: Oops, got a little too excited cleaning everything up!

src/mdio/converters/segy.py
```diff
@@ -394,13 +394,19 @@ def segy_to_mdio(  # noqa: PLR0913, PLR0915
...
+    valid_count = int(np.count_nonzero(valid_mask))
+    if valid_count != num_traces:
         for dim_name in grid.dim_names:
-            dim_min, dim_max = grid.get_min(dim_name), grid.get_max(dim_name)
+            dim_min = grid.get_min(dim_name)
+            dim_max = grid.get_max(dim_name)
             logger.warning("%s min: %s max: %s", dim_name, dim_min, dim_max)
         logger.warning("Ingestion grid shape: %s.", grid.shape)
-        raise GridTraceCountError(np.sum(grid.live_mask), num_traces)
+        raise GridTraceCountError(valid_count, num_traces)
```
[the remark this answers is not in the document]
BrianMichell: It does use more memory in the main process compared to what is live in 0.9.2 (below), but it has a lower peak memory consumption. This should only account for a few megabytes though, and I'll look into any other low-hanging fruit as far as memory consumption.

image

src/mdio/core/grid.py
tasansal: hm, if i understood correctly, if we are writing the live mask from distributed tasks (which are chunked same as seismic) it will cause race conditions and corrupt the live mask (which has larger chunks). e.g. multiple workers may write to the same chunk at the same time. Have you checked if live mask is identical before/after this?

BrianMichell: We are not writing the live mask in a distributed manner, and the Grid.live_mask chunking also differs from the live mask which is written to durable media. I believe this is 100% safe behavior.

src/mdio/converters/segy.py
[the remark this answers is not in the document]
BrianMichell: I've validated that it only reaches this block once per live mask chunk. For a 100GB file, this is once. Reduced the maximum size of the live mask and saw the appropriate amount of iterations.

PR #577, opened 2025-07-26
segy_to_mdio_v1

dmitriyrepin:
**This is a draft PR**
Branch: 
* https://github.com/dmitriyrepin/mdio-python/tree/segy_to_mdio_v1
Any critique or code change suggestions are highly appreciated.

# Major changes:
* I have used Altay's approach for [src/mdio/converter/segy_to_mdio_1.py](https://github.com/dmitriyrepin/mdio-python/blob/segy_to_mdio_v1/src/mdio/converters/segy_to_mdio_v1.py) using his factory code as a reference. I have refactored it a bit into functions to make it easier for me to comprehend. I added a lot of comments that can be removed in the final version.

* I have updated AbstractDatasetTemplate to [add trace_mask and trace_headers](https://github.com/dmitriyrepin/mdio-python/blob/340f78a5cf424b2e49d5052ad9dc33b33918091b/src/mdio/schemas/v1/templates/abstract_dataset_template.py#L86) variables to the dataset created in [build_dataset](https://github.com/dmitriyrepin/mdio-python/blob/340f78a5cf424b2e49d5052ad9dc33b33918091b/src/mdio/schemas/v1/templates/abstract_dataset_template.py#L66C9-L66C22), which now takes headers: StructuredType to specify the headers type
  * I am still debugging the issue with trace header writing, but now I face a different error.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -41,6 +41,6 @@
   "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/mdio-python,type=bind",
   "workspaceFolder": "/workspaces/mdio-python",
   "mounts": [
-    "source=${localWorkspaceFolder}/../DATA/,target=/DATA/,type=bind,consistency=cached"
+    // "source=${localWorkspaceFolder}/../DATA/,target=/DATA/,type=bind,consistency=cached"
   ]
 }


src/mdio/schemas/v1/dataset_serializer.py
```diff
@@ -294,4 +293,5 @@ def to_zarr(
     """
     kwargs["zarr_format"] = 2
     kwargs["compute"] = False
+    kwargs["mode"] = "w"  # create (overwrite if exists)
```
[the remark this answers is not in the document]
dmitriyrepin: I see no value for this method. 
I think, it should be removed altogether in favour of directly using Xarray's to_zar.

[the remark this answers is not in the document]
BrianMichell: The purpose of this convenience function is to abstract away as much of the xarray/zarr backend as possible. This should enable us to provide drop-in backend replacements that don't require changes to client code (like upgrading to use Zarr  v3 encoding).

tasansal:
>   I believe Dask does not support StructuredType data type. I will need to talk to you how to work around of this issue. I have a couple of creasy suggestions.

Dask does support numpy data types which are derived from StructuredType.

src/mdio/core/indexing_v1.py
```diff
@@ -0,0 +1,102 @@
+"""Indexing logic."""
+
+import itertools
+from math import ceil
+
+import numpy as np
+from zarr import Array
+
+class ShapeAndChunks:
```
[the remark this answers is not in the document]
dmitriyrepin: I can remove this. 

I was trying to make ChunkIterator() least demanding.
```Python
class ShapeAndChunks:
    """A class that has the .shape and .chunks properties
    It can be duck-typed by any other type that has .shape and .chunks properties.
    For example, zarr.Array
```
I thought that signature:
def __init__(self, shape_chunks: ShapeAndChunks, dims: list[str], chunk_samples: bool = True):
would better indicate that both zarr, dask can be passed in.

While, it seemed to me, that the signature
from zarr import Array
def __init__(self, array: Array, dims: list[str], chunk_samples: bool = True):
indicates to the user that only zarr arrays should be passed and not the dask arrays. 

However, as you properly mention, in python one still can pass the dask array

src/mdio/converters/type_converter.py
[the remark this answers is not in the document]
dmitriyrepin: Here is the reason why I thought the converter is needed, but I might be mistaken:
The dataset builder expects a StructuredType
```
    def add_variable(  # noqa: PLR0913
        self,
        name: str,
        *,
        long_name: str = None,
        dimensions: list[str],
        data_type: ScalarType | StructuredType,
        compressor: Blosc | ZFP | None = None,
        coordinates: list[str] | None = None,
        metadata_info: VariableMetadataList | None = None,
    )
```
for exmaple
```
StructuredType(
        fields=[
                StructuredField(name="cdp_x", format=ScalarType.INT32),
                StructuredField(name="cdp_y", format=ScalarType.INT32),
                StructuredField(name="elevation", format=ScalarType.FLOAT16),
                StructuredField(name="some_scalar", format=ScalarType.FLOAT16),
        ]
    )
```
Scanning SegY file results in index_headers: list[Dimensions] that resolves to:
```
np.dtype([('cdp_x', '<i4'), ('cdp_y', '<i4'), ('elevation', '<f2'), ('some_scalar', '<f2')])
```
So, somehow one needs to convert the np.dtype to StructuredType when building a Pydentic model before populating it with data from SEG-Y, unless the StructuredType is fixed to have only SEG_Y headers and no custom or v1 headers
 
So I wrote to_structured_type() for the conversion. It could have been very simple as to_numpy_dtype().
However, the numpy uses both "float32" and "<f4" to encode the same type and this make the converter so complex.

src/mdio/core/indexing_v1.py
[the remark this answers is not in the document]
dmitriyrepin: Here is also gotcha: xarray does support .shape and .chunks, but chunks are not what I expected 
None, if xr_array.chunk() was not called or otherwise
```
xr_array.chunks
((2, 2, 2, 2), (2, 2, 2, 2), (2,))
```
Instead it requires:
```
     shape=data.shape,
     chunks= data.encoding.get("chunks"),
```
Maybe we could keep the class, but add a convenience function to create it from different array types?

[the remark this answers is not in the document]
dmitriyrepin: Removed from indexing

src/mdio/converters/type_converter.py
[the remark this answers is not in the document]
dmitriyrepin: to_structured_type() could have been very simple as to_numpy_dtype().
However, the numpy uses both "float32" and "<f4" to encode the same type and this makes the converter so complex.

src/mdio/converters/segy_to_mdio_v1.py
```diff
@@ -0,0 +1,241 @@
...
+    grid.build_map(index_headers)
+
+    # Create grid map and trace mask arrays
+    grid_map, trace_mask = _create_grid_map_trace_mask(grid=grid)
+    xr_sd.trace_mask.data[:] = trace_mask
+
+    # Populate the "trace_mask" variable in the xarray dataset and write it to Zarr
+    # Get subset of the dataset containing only "trace_mask"
+    #TODO: Should we write the trace mask inside of blocked_io.to_zarr?
+    #TODO: Should we clear the coords from ds_to_write?
```
tasansal, line 163: i think both are ok the way it is now in your code. no need to write these in blocked_io or clear coords; they're usually small.

src/mdio/converters/segy_to_mdio_v1.py
```diff
@@ -0,0 +1,241 @@
...
+                                                         coordinates=coordinates,
+                                                         output=output)
+
+    grid_map = _populate_trace_mask_and_write_to_zarr(segy_file=segy_file,
+                                                      dimensions=dimensions,
+                                                      index_headers=index_headers,
+                                                      xr_sd=xr_dataset,
+                                                      output=output)
+
+    # TODO: Maybe we should have saved the data_variable_name as a Dataset attribute?
```
tasansal, line 227: yea we should do that, we need to update the schema if we want to enforce it. DatasetMetadata can have a field for "default_variable" that stores this value.

src/mdio/converters/segy_to_mdio_v1_custom.py
tasansal, line 1: let's remove this module, we don't need it at the moment, and we haven't spec'd out what the API should be like. we can leave it for another task.

src/mdio/converters/type_converter.py
```diff
@@ -0,0 +1,114 @@
+from numpy import dtype as np_dtype
+from mdio.schemas.dtype import ScalarType
+from mdio.schemas.dtype import StructuredType
+from mdio.schemas.dtype import StructuredField
+from mdio.schemas.v1.dataset_builder import _to_dictionary
+from segy.arrays import HeaderArray as HeaderArray
+
+def _to_scalar_type_using_kind_and_size(field_dtype: np_dtype, field_name: str) -> ScalarType:
```
tasansal, line 8: this can be simplified significantly. if you build a `np.dtype` from a `field_dtype` no matter what form it is; it gives an instance of a `np.dtype` that can extract EXACT name so we don't have to enumerate every single option.

e.g. all of the below evaluate to `"int32"` which can be directly requested from `ScalarType` by value.

```python
>>> dt = np.dtype("<i4")
>>> dt.name
"int32"
```

```python
>>> dt = np.dtype(">i4")
>>> dt.name
"int32"
```

```python
>>> dt = np.dtype("int32")
>>> dt.name
"int32"
```

```python
>>> ScalarType(dt.name)
<ScalarType.INT32: 'int32'>
```

src/mdio/converters/type_converter.py
```diff
@@ -0,0 +1,114 @@
...
+        elif field_dtype.itemsize == 16:
+            return ScalarType.COMPLEX128
+        else:
+            raise ValueError(f"Unsupported complex size: {field_dtype.itemsize} bytes for field '{field_name}'")
+    elif field_dtype.kind == 'b':  # boolean
+        return ScalarType.BOOL
+    else:
+        raise ValueError(f"Unsupported numpy dtype: {field_dtype} for field '{field_name}'")
+
+def to_scalar_type(field_dtype: np_dtype, field_name: str) -> ScalarType:
```
tasansal, line 53: see comment above

src/mdio/core/grid.py
```diff
@@ -106,47 +107,62 @@ def from_zarr(cls, zarr_root: zarr.Group) -> Grid:
...
+        for dim_idx, sl in enumerate(chunk_slices):
+            arr = self.header_index_arrays[dim_idx]  # shape: (num_traces,)
+            start, stop = sl.start, sl.stop
+            if start is not None:
+                mask &= arr >= start
+            if stop is not None:
+                mask &= arr < stop
+            if not mask.any():
+                # No traces remain after this dimension's filtering
+                return np.empty((0,), dtype=np.uint32)
```
tasansal, line 162: what is going on here? what logic does it replace/add ?

src/mdio/core/indexing.py
```diff
@@ -8,87 +8,76 @@
...
-        if self._idx < self.num_chunks:
+    def __next__(self) -> dict[str, slice]:
+        """Iteration logic."""
+        if self._idx <= self.num_chunks:
+            # We build slices here. It is dimension agnostic
             current_start = next(self._ranges)
 
             start_indices = tuple(
-                dim * chunk for dim, chunk in zip(current_start, self.len_chunks, strict=True)
+                dim * chunk for dim, chunk in zip(current_start, self.len_chunks)
```
tasansal, line 66: the removal of `strict` from the `zip` iterators will yield in linting errors, please add them back.

src/mdio/factory_v1.py
tasansal, line 1: this needs to go, the old implementation

src/mdio/segy/_workers.py
```diff
@@ -51,71 +54,47 @@ def header_scan_worker(segy_file: SegyFile, trace_range: tuple[int, int]) -> Hea
...
-
-    tmp_data = np.zeros(seq_trace_indices.shape + (grid.shape[-1],), dtype=data_array.dtype)
-    tmp_metadata = np.zeros(seq_trace_indices.shape, dtype=metadata_array.dtype)
+    live_trace_indexes = grid_map[not_null].tolist()
+    traces = segy_file.trace[live_trace_indexes]
 
-    del grid  # To save some memory
+    # Get subset of the dataset that has not yet been saved
+    # The headers might not be present in the dataset
+    if "headers" in dataset.data_vars:
```
tasansal, line 76: instead of condoitional logic here, we should just apply the same logic to all data variables that match the dimensions of the grid exactly. no need to treat every option separately. e.g. what happens if there is "headers", and 4 more data variables in the schema.

tasansal (changes requested):
Hey, looks much better! some more changes and we should be good to go.

src/mdio/converters/type_converter.py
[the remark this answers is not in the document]
dmitriyrepin: I have simplified the `to_scalar_type` and `to_structured_type` and documented their limitations

src/mdio/segy/blocked_io_v1.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # Get subset of the dataset that has not yet been saved
+    # The headers might not be present in the dataset
+    if "headers" in dataset.data_vars:
+        ds_to_write = dataset[[data_variable_name, "headers"]]
+        ds_to_write = dataset.reset_coords()
+        ds_to_write = dataset.drop_vars(["trace_mask"])
+        ds_to_write["headers"].data[not_null] = traces.header
+        ds_to_write["headers"].data[~not_null] = 0
+        # BUG: Fails here with "IndexError: Boolean array with size 400 is not long enough for axis 0 with size 20"At the moment
```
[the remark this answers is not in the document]
dmitriyrepin: I moved the trace_worker function to _workers.py
I can't observe the bug any more, but not sure how it was fixed, probably because I have removed an erroneous xr_Dataset.chunk() call somewhere in the code.

src/mdio/converters/segy_to_mdio_v1_custom.py
dmitriyrepin: I have removed the module, but kept the segy_to_mdio_v1_customized function that allows to pass in index_bytes and index_names to customize the SegySpec before calling segy_to_mdio_v1. 
We need this because a lot of test datasets use customization

src/mdio/converters/type_converter.py
dmitriyrepin: I accidently pushed an experimental version of the grid. I have reverted the change back

src/mdio/core/indexing.py
dmitriyrepin: Done

src/mdio/segy/_workers.py
dmitriyrepin: The conditional logic there is to handle the case of the datasets that do not have headers

src/mdio/core/grid.py
```diff
@@ -106,47 +107,62 @@ def from_zarr(cls, zarr_root: zarr.Group) -> Grid:
         return cls(dims_list)
 
     def build_map(self, index_headers: HeaderArray) -> None:
-        """Build trace mapping and live mask from header indices.
+        """Compute per-trace grid coordinates (lazy map).
+
+        Instead of allocating a full `self.map` and `self.live_mask`, this computes, for each trace,
```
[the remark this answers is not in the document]
dmitriyrepin: I accidently pushed an experimental version of the grid. I have reverted the change back

src/mdio/factory_v1.py
dmitriyrepin: Removed

BrianMichell:
[FYI](https://github.com/pydata/xarray/discussions/10587) I'll keep investigating the structured data fill value issue.

src/mdio/segy/blocked_io.py
```diff
@@ -14,100 +14,148 @@
...
-            chunker,
-            chunksize=pool_chunksize,
-        )
-
-        lazy_work = tqdm(
-            iterable=lazy_work,
+        futures = []
+        common_args = (segy_file, out_path, data_variable_name)
+        for region in chunk_iter:
+            index_slices = tuple(region[key] for key in data.dims[:-1])
```
BrianMichell, line 125: I don't think we need to explicitly drop the last dim here.

src/mdio/segy/blocked_io.py
```diff
@@ -14,100 +14,148 @@
...
+
+def _update_stats(final_stats: SummaryStatistics, partial_stats: SummaryStatistics) -> None:
+    final_stats.count += partial_stats.count
+    final_stats.min = min(final_stats.min, partial_stats.min)
+    final_stats.max = min(final_stats.max, partial_stats.max)
+    final_stats.sum += partial_stats.sum
+    final_stats.sum_squares += partial_stats.sum_squares
+
+
+class ShapeAndChunks:
```
BrianMichell, line 54: Is it necessary to have a full blown class for this?

pyproject.toml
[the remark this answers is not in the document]
dmitriyrepin: The `jupyter` and `matplotlib` are not the direct MDIO dependencies. So, installing MDIO via 'pip install multidimio` will not install them.

However, the project includes both Jupyter notebook files in the documentation and examples of plotting with matplotlib. Thus, I included them as developer's dependencies (which can be installed via 'uv sync --group dev') to allows setup of a complete working development environment (see .devcontainer/Dockerfile.dev).

src/mdio/segy/_workers.py
```diff
@@ -51,71 +53,46 @@ def header_scan_worker(segy_file: SegyFile, trace_range: tuple[int, int]) -> Hea
...
-    live_subset = grid.live_mask[chunk_indices[:-1]]
-
-    if np.count_nonzero(live_subset) == 0:
+    out_path: str,
+    data_variable_name: str,
+    region: dict[str, slice],
+    grid_map: zarr_Array,
+    dataset: xr_Dataset,
+) -> SummaryStatistics | None:
+    """Read a subset of traces and write to region of Zarr file."""
```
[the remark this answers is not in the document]
dmitriyrepin: Fixed. I did run pre-commit, but it somehow did not complain here.

src/mdio/converters/segy_to_mdio_v1.py
```diff
@@ -0,0 +1,298 @@
...
+    grid_chunksize = None
+    segy_dimensions, chunksize, index_headers = get_grid_plan(
+        segy_file=segy_file, return_headers=True, chunksize=grid_chunksize, grid_overrides=None
+    )
+    return segy_dimensions, index_headers
+
+
+def _get_data_coordinates(
+    segy_headers: list[Dimension], mdio_template: AbstractDatasetTemplate
+) -> tuple[list[Dimension], list[Dimension]]:
```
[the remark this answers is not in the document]
dmitriyrepin: Added Args and Returns to the docstring. I hope, it clarifies it now.

src/mdio/core/indexing.py
```diff
@@ -8,72 +8,58 @@
...
+        # Handle the case when we don't want to slice the sample axis.
+        if chunk_samples is False:
             self.len_chunks = self.len_chunks[:-1] + (self.arr_shape[-1],)
 
-        # Calculate the number of chunks per dimension
+        # Compute number of chunks per dimension, and total number of chunks
         self.dim_chunks = [
             ceil(len_dim / chunk)
-            for len_dim, chunk in zip(self.arr_shape, self.len_chunks, strict=True)
+            for len_dim, chunk in zip(self.arr_shape, self.len_chunks, strict=False)
```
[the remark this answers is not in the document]
dmitriyrepin: Typo? Fixed

src/mdio/converters/segy_to_mdio_v1.py
```diff
@@ -0,0 +1,298 @@
...
+                dataset.drop_vars(c.name)
+                continue
+        # Otherwise, populate the coordinate variable
+        values = c.coords
+        shape = dataset.coords[c.name].shape
+        dims = dataset.coords[c.name].dims
+        if shape != values.shape:
+            values = values.reshape(shape)
+        dataset.coords[c.name] = (dims, values)
+        vars_to_drop_later.append(c.name)
```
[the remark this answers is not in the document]
dmitriyrepin: Hmm... I do not believe there is a side-effect here:
* dataset.coords[c.name] comes in as N-dimensional
* the input data c.coords always come as a flattened 1D array that needs to be reshaped to the proper N-dims,
but dimensionality of dataset.coords[c.name] does not change.

src/mdio/schemas/v1/templates/abstract_dataset_template.py
```diff
@@ -131,13 +151,56 @@ def _add_coordinates(self) -> None:
                 metadata_info=[self._coord_units],
             )
 
+    def _add_trace_mask(self) -> None:
```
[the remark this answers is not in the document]
BrianMichell: Relevant snippet (from https://github.com/TGSAI/mdio-python/pull/527/files#diff-8f0a1f055ed04b15d121e3c9c6fb4ea8e72f15a7fbd56a9bc551b665f16f6045R91)
```python
        grid_size = np.prod(self.shape[:-1])
        map_dtype = "uint64" if grid_size > UINT32_MAX else "uint32"
        fill_value = np.iinfo(map_dtype).max

        # Initialize Zarr arrays for the map and live mask
        live_shape = self.shape[:-1]
        chunks = get_constrained_chunksize(
            shape=live_shape,
            dtype=map_dtype,
            max_bytes=self._INTERNAL_CHUNK_SIZE_TARGET,
        )
        # Temporary zarrs for ingestion.
        self.map = zarr.full(live_shape, fill_value, dtype=map_dtype, chunks=chunks)
        self.live_mask = zarr.zeros(live_shape, dtype="bool", chunks=chunks)

        # Calculate batch size for processing
        memory_per_trace_index = index_headers.itemsize
        batch_size = int(self._TARGET_MEMORY_PER_BATCH / memory_per_trace_index)
        total_live_traces = index_headers.size
```

src/mdio/converters/segy_to_mdio_v1.py
```diff
@@ -0,0 +1,297 @@
...
+
+@dataclass
+class StorageLocation:
+    """A class to represent a local or cloud storage location for SEG-Y or MDIO files."""
+
+    uri: str
+    options: dict[str, Any] = field(default_factory=dict)
+
+    @property
+    def storage_type(self) -> str:
```
BrianMichell, line 44: I don't think we want to maintain the `storage_type` and `exists` methods. These should all be abstracted by `fsspec`.

src/mdio/converters/segy_to_mdio_v1.py
```diff
@@ -0,0 +1,297 @@
...
+        shape = dataset.coords[c.name].shape
+        dims = dataset.coords[c.name].dims
+        if shape != values.shape:
+            values = values.reshape(shape)
+        dataset.coords[c.name] = (dims, values)
+        vars_to_drop_later.append(c.name)
+    return dataset
+
+
+def _get_horizontal_coordinate_unite(segy_headers: list[Dimension]) -> LengthUnitEnum | None:
```
BrianMichell, line 152:
```suggestion
def _get_horizontal_coordinate_unit(segy_headers: list[Dimension]) -> LengthUnitEnum | None:
```

src/mdio/converters/segy_to_mdio_v1.py
```diff
@@ -0,0 +1,297 @@
...
+
+    # Scan the SEG-Y file for headers
+    # This is an memory-expensive and time-consuming read-write operation
+    segy_headers, segy_index_headers = _scan_for_headers(segy_file)
+    # Extract dim and non-dim coordinates according to the MDIO template
+    dimension_coords, non_dim_coords = _get_data_coordinates(segy_headers, mdio_template)
+    shape = [len(dim.coords) for dim in dimension_coords]
+    headers = to_structured_type(segy_index_headers.dtype)
+
+    horizontal_unit = _get_horizontal_coordinate_unite(segy_headers)
```
BrianMichell, line 261:
```suggestion
    horizontal_unit = _get_horizontal_coordinate_unit(segy_headers)
```

src/mdio/segy/_workers.py
```diff
@@ -51,71 +53,60 @@ def header_scan_worker(segy_file: SegyFile, trace_range: tuple[int, int]) -> Hea
...
 
     Returns:
-        Partial statistics for chunk, or None
+        SummaryStatistics object containing statistics about the written traces.
     """
-    # Special case where there are no traces inside chunk.
-    live_subset = grid.live_mask[chunk_indices[:-1]]
-
-    if np.count_nonzero(live_subset) == 0:
+    if dataset.trace_mask.sum() == 0:
```
BrianMichell, line 77:
```suggestion
    if not dataset.trace_mask.any():
```
I believe this will be a more efficient operation. Could you please double check this assumption?

tests/integration/v1/test_segy_to_mdio_v1.py
```diff
@@ -0,0 +1,352 @@
...
+from mdio.converters.segy_to_mdio_v1 import StorageLocation
+from mdio.converters.segy_to_mdio_v1 import segy_to_mdio_v1
+from mdio.converters.type_converter import to_numpy_dtype
+from mdio.schemas.dtype import ScalarType
+from mdio.schemas.dtype import StructuredField
+from mdio.schemas.dtype import StructuredType
+from mdio.schemas.v1.templates.template_registry import TemplateRegistry
+
+
+def _slice_three_values(dims: tuple[int], values_from_start: bool) -> Any:
```
BrianMichell, line 21: Style nit-pick: Directly return from conditionals here.

BrianMichell (changes requested):
We're getting closer.

We may want to reduce the scope of these tasks in the future, it's pretty unwieldly to review 2,500+ lines of changes repeatedly.

src/mdio/core/indexing.py
```diff
@@ -8,72 +8,58 @@
...
+
+    Attributes:             # noqa: DOC602
+        arr_shape: Shape of the array.
+        len_chunks: Length of chunks in each dimension.
+        dim_chunks: Number of chunks in each dimension.
+        num_chunks: Total number of chunks.
     """
 
-    def __init__(self, array: Array, chunk_samples: bool = True):
+    def __init__(self, array: Array, dims: list[str], chunk_samples: bool = True):
```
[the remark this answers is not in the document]
dmitriyrepin: I have refactored the ChunkIterator into ChunkIteratorV1.
It now supports both the slice dictionary and the slice tuple returned while iterating.
The slice tuple can continued to be used in src/mdio/api/convenience.py
The slice dictionary can now be used in the blocked_io.to_zarr_v1 with dataset region selection (dataset.isel(region)
Example of usage:
        >> chunks = (3, 4, 5)
        >> shape = (5, 11, 19)
        >> dims = ["inline", "crossline", "depth"]
        >>
        >> iter = ChunkIteratorV1(shape=shape, chunks=chunks, dim_names=dims)
        >> for i in range(13):
        >>    region = iter.__next__()
        >> print(region)
        { "inline": slice(3,6, None), "crossline": slice(0,4, None), "depth": slice(0,5, None) }
 
        >> iter = ChunkIteratorV1(shape=shape, chunks=chunks, dim_names=None)
        >> for i in range(13):
        >>    region = iter.__next__()
        >> print(region)
        (slice(3,6,None), slice(0,4,None), slice(0,5,None))

Please comment and critique.

src/mdio/segy/_workers.py
```diff
@@ -51,71 +53,58 @@ def header_scan_worker(segy_file: SegyFile, trace_range: tuple[int, int]) -> Hea
...
-    tmp_metadata[live_subset] = headers.view(tmp_metadata.dtype)
-    tmp_data[live_subset] = samples
+    live_trace_indexes = grid_map[not_null].tolist()
+    traces = segy_file.trace[live_trace_indexes]
 
-    # Flush metadata to zarr
-    metadata_array.set_basic_selection(selection=chunk_indices[:-1], value=tmp_metadata)
+    # Get subset of the dataset that has not yet been saved
+    # The headers might not be present in the dataset
+    # Q: to save memory, should we overwrite the 'dataset'?
```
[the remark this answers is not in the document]
dmitriyrepin: Done: https://github.com/TGSAI/mdio-python/issues/584

src/mdio/segy/blocked_io.py
```diff
@@ -14,100 +14,148 @@
...
+        for future in iterable:
+            result = future.result()
+            if result is not None:
+                _update_stats(final_stats, result)
+
+    # Xarray doesn't directly support incremental attribute updates when appending to an
+    # existing Zarr store.
+    # HACK: We will update the array attribute using zarr's API directly.
+    # Open the Zarr store using zarr directly
+    zarr_group = zarr_open_group(f"{out_path}", mode="a")
```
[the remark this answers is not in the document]
dmitriyrepin: Done

src/mdio/segy/blocked_io.py
```diff
@@ -14,100 +14,148 @@
...
+    attr_json = final_stats.model_dump_json()
+    # Use the data_variable_name to get the array in the Zarr group
+    # and write "statistics" metadata there
+    zarr_group[data_variable_name].attrs.update({"statistics": attr_json})
```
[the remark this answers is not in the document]
dmitriyrepin: Done

tests/integration/v1/test_segy_to_mdio_v1.py
dmitriyrepin: Since this is a test code, didn't want to create a test for a test. :-)
So, I had to have an intermediate value `slices` available to check in the debugger.

src/mdio/converters/segy_to_mdio_v1.py
dmitriyrepin: Thank you. Fixed.

src/mdio/converters/segy_to_mdio_v1.py
```diff
@@ -0,0 +1,298 @@
...
+
+    dataset.to_zarr(
+        store=output.uri, mode="w", write_empty_chunks=False, zarr_format=2, compute=True
+    )
+
+    # Now we can drop them to simplify chunked write of the data variable
+    return dataset.drop_vars(vars_to_drop_later)
+
+
+def _populate_trace_mask_write_to_zarr(
```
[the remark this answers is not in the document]
dmitriyrepin: Done

src/mdio/schemas/v1/templates/abstract_dataset_template.py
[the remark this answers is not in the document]
dmitriyrepin: I have created an enhancement issue "Enhance segy_to_mdio_v1 to do chunked write for non-dimensional coordinates and trace_mask #587"
https://github.com/TGSAI/mdio-python/issues/587

src/mdio/schemas/v1/templates/abstract_dataset_template.py
```diff
@@ -111,18 +151,30 @@ def _load_dataset_attributes(self) -> UserAttributes:
...
-        """A virtual method that can be overwritten by subclasses to add custom dimensions.
+        """Add custom dimensions.
 
+        A virtual method that can be overwritten by subclasses to add custom dimensions.
         Uses the class field 'builder' to add dimensions to the dataset.
         """
         for i in range(len(self._dim_names)):
             self._builder.add_dimension(self._dim_names[i], self._dim_sizes[i])
 
     def _add_coordinates(self) -> None:
```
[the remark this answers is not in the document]
dmitriyrepin: I have created an enhancement issue "Enhance segy_to_mdio_v1 to do chunked write for non-dimensional coordinates and trace_mask #587"
https://github.com/TGSAI/mdio-python/issues/587

src/mdio/converters/segy_to_mdio_v1.py
```diff
@@ -0,0 +1,298 @@
...
+    data_variable_name = mdio_template.trace_variable_name
+    # This is an memory-expensive and time-consuming read-write operation
+    # performed in chunks to save the memory
+    blocked_io.to_zarr(
+        segy_file=segy_file,
+        out_path=output_location.uri,
+        grid_map=grid_map,
+        dataset=xr_dataset,
+        data_variable_name=data_variable_name,
+    )
```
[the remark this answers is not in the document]
dmitriyrepin: We are writing statistics within blocked_io.to_zarr_v1().
Since  Xarray doesn't directly support incremental attribute updates when appending to an existing Zarr store,
I update the array attribute using zarr's API directly.

src/mdio/converters/segy_to_mdio_v1.py
dmitriyrepin: Updated StorageLocation to use fsspec to check existance.
Since now trace_worker_v1 depends on it, I moved it to src/mdio/core/storage_location.py
Please note that the class design has changed from data class to a normal class to ensure that the uri and the options to be read-only immutable properties.

src/mdio/core/storage_location.py
```diff
@@ -0,0 +1,85 @@
+"""StorageLocation class for managing local and cloud storage locations."""
+
+from pathlib import Path
+from typing import Any
+
+import fsspec
+
+
+class StorageLocation:
```
BrianMichell, line 9: I'm wondering if there's a more elegant way we can lean on `fsspec` here. We also should be supporting Azure Blob Storage, and it would be exceptionally helpful to allow other storage drivers such as FTP, less used cloud storage providers, or OSDU implementations to use StorageLocation without requiring code changes on our end.

This can be deferred if it's not trivial.

@tasansal Thoughts?

src/mdio/schemas/v1/dataset_serializer.py
```diff
@@ -228,12 +218,14 @@ def to_xarray_dataset(mdio_ds: Dataset) -> xr_DataArray:  # noqa: PLR0912
         # Create a custom chunk key encoding with "/" as separator
         chunk_key_encoding = V2ChunkKeyEncoding(separator="/").to_dict()
         encoding = {
-            # NOTE: See Zarr documentation on use of fill_value and _FillValue in Zarr v2 vs v3
-            "_FillValue": _get_fill_value(v.data_type),
             "chunks": chunks,
             "chunk_key_encoding": chunk_key_encoding,
             "compressor": _convert_compressor(v.compressor),
         }
+        if v.name not in no_fill_var_names:
```
BrianMichell, line 225:
```suggestion
        if dtype.fields is not None:
```

I think this will allow us to get away the fill value issue without requiring us to pass `no_fill_var_names`, unless there was another reason to allow omitting fill values.

BrianMichell (approved):
A few minor things, but nothing I see blocking us from progressing at this point.

tasansal:
@dmitriyrepin I fixed some critical bugs and erroneous logic. In the meantime updated the incorrect tests to check for correct data types. The major problem was that the Xarray Dataset dimension coordinates were not populated with the proper method, and it was type-casting in undefined ways. The tests validating the variables were using the wrong data type. I fixed them but now one of the tests still have an issue. Will check that tomorrow morning.

tasansal:
@dmitriyrepin there is a bug here, potentially some encoding properties missing:

https://github.com/TGSAI/mdio-python/blob/0d410d35c02402fc71839778a7a0cebe1a72665c/src/mdio/converters/segy_v1.py#L251-L253

Everything looks good on the in-memory representation of xarray (dtypes for dimension coordinates are int32 as expected). But if you open the file right after this call; they turn into float64.

PR #599, opened 2025-08-09
Fix integration tests for segy_to_dio import

dmitriyrepin:
* Fix structured types bug
* Fix templates to use the proper trace header names
* Add Seismic2DPreStackCDPTemplate and Seismic2DPreStackShotTemplate 
* Fix test_impoort
image

### DIFF
--- a/tests/integration/test_segy_import_export_masked.py
+++ b/tests/integration/test_segy_import_export_masked.py
@@ -8,20 +8,25 @@ expected full or partial files.
 from __future__ import annotations
 
 from dataclasses import dataclass
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 import fsspec
 import numpy as np
 import pytest
+import xarray as xr
 from numpy.testing import assert_array_equal
 from segy import SegyFile
 from segy.factory import SegyFactory
 from segy.schema import HeaderField
+from segy.schema import SegySpec
 from segy.standards import get_segy_standard
 
 from mdio import MDIOReader
 from mdio import mdio_to_segy
-from mdio import segy_to_mdio
+from mdio.converters.segy import segy_to_mdio
+from mdio.core.storage_location import StorageLocation
+from mdio.schemas.v1.templates.template_registry import TemplateRegistry
 from mdio.segy.utilities import segy_export_rechunker
 
 if TYPE_CHECKING:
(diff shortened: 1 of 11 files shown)


tests/integration/test_segy_import_export_masked.py
```diff
@@ -192,10 +201,24 @@ def mock_nd_segy(path: str, grid_conf: GridConfig, segy_factory_conf: SegyFactor
...
     headers["coord_scalar"] = -100
-    for field in ["cdp_x", "src_x", "rec_x"]:
-        headers[field] = np.random.randint(low=700000, high=900000, size=trace_numbers.size) * 100
-    for field in ["cdp_y", "src_y", "rec_y"]:
-        headers[field] = np.random.randint(low=4000000, high=5000000, size=trace_numbers.size) * 100
+    for field in ["cdp_x", "source_coord_x", "group_coord_x"]:
+        start = 700000
+        step = 100
+        stop = start + step * (trace_numbers.size - 0)
+        headers[field] = np.arange(start=start, stop=stop, step=step)
```
BrianMichell, line 208: What was the motivation going from random to ordered values here? It may fit the nominal case, but is un-sorted data within the scope of these tests as well?

tests/integration/test_segy_import_export_masked.py
```diff
@@ -192,10 +201,24 @@ def mock_nd_segy(path: str, grid_conf: GridConfig, segy_factory_conf: SegyFactor
...
+    for field in ["cdp_y", "source_coord_y", "group_coord_y"]:
+        start = 4000000
+        step = 100
+        stop = start + step * (trace_numbers.size - 0)
+        headers[field] = np.arange(start=start, stop=stop, step=step)
+
+    # Array filled with repeating sequence
+    sequence = np.array([1, 2, 3])
+    # Calculate the number of times the sequence needs to be repeated
+    aaa = trace_numbers.size / len(sequence)
```
BrianMichell, line 218: Rename `aaa` please.

tests/integration/test_segy_import_export_masked.py
```diff
@@ -266,31 +307,52 @@ def test_ingested_mdio(self, test_conf: MaskedExportConfig, export_masked_path:
...
+        for expected in expected_dims:
+            actual_dim = ds[expected.name]
+            assert expected.name == actual_dim.name
+            assert expected.size == actual_dim.values.size
+            assert expected.start == actual_dim.values[0]
 
-        # Read all the data into memory
-        live, headers, traces = actual_reader[..., 0]
+        live_mask = ds["trace_mask"].values
+        num_traces = np.prod(live_mask.shape)
```
BrianMichell, line 324: We want `np.sum` here, don't we? If we just want the hyper-rectangle maybe we should rename to `candidate_traces` or similar to avoid confusing the term?

BrianMichell (changes requested):
Only a few things stood out to me.

tests/integration/test_segy_import_export_masked.py
dmitriyrepin: The intension was to access a 2D bool array of the live trace mask. 
In the test case it has been populated with all True, which is validated below.
The total number of traces in the 3D survey - width * height of live_mask (num_traces = np.prod(live_mask.shape))
I am not sure to what the `candidate_traces` term would refer to.

src/mdio/core/grid.py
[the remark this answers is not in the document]
dmitriyrepin: src/mdio/core/grid.py:154
Why do we try to size it as np.uint32 in the first place, when we know it will be resized as np.uint64 after
```Python
self.map.vindex[live_dim_indices] = trace_indices
```

tests/integration/test_segy_import_export_masked.py
BrianMichell: Since we are accessing the `live_mask` we should assume raggedness, even if we know that the test is dense and perfectly rectangular. `np.prod` insinuates (to me) that we are more interested in the potential volume of the array rather than the contents.

tasansal (approved):

PR #600, opened 2025-08-11
Serialize text and binary headers

### LINKED ISSUE #595: Serialize text and binary headers to MDIO attributes
Currently v1 doesn't save this information. We need to save it somewhere.

dmitriyrepin:
Closes #595 

Address Serialize text and binary headers to MDIO attributes
Serialize text and binary headers to MDIO attributes


tasansal (changes requested):
Looks good to me, we just need to add some tests. two options:

use the existing teapot (tgsai/segy can open remote segy in place without downloading) to check; or use the mock segy files we generate to compare. ideally both; but I would be ok with mock segy tests for binary/text header first to merge this PR. we can do the teapot if needed later.

dmitriyrepin:
I have added the teapod text and binary header test

PR #607, opened 2025-08-15
Enable PreStackCocaGathers3D integration tests

dmitriyrepin:
* Updated the Coca template no-dimensional variables to use all 4 coordinates 
* Enable PreStackCocaGathers3D integration tests

### DIFF
--- a/src/mdio/schemas/v1/templates/seismic_3d_prestack_coca.py
+++ b/src/mdio/schemas/v1/templates/seismic_3d_prestack_coca.py
@@ -64,13 +64,13 @@ class Seismic3DPreStackCocaTemplate(AbstractDatasetTemplate):
         # Add non-dimension coordinates
         self._builder.add_coordinate(
             "cdp_x",
-            dimensions=["inline", "crossline"],
+            dimensions=["inline", "crossline", "offset", "azimuth"],
             data_type=ScalarType.FLOAT64,
             metadata_info=[self._horizontal_coord_unit],
         )
         self._builder.add_coordinate(
             "cdp_y",
-            dimensions=["inline", "crossline"],
+            dimensions=["inline", "crossline", "offset", "azimuth"],
             data_type=ScalarType.FLOAT64,
             metadata_info=[self._horizontal_coord_unit],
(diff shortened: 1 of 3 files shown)


src/mdio/schemas/v1/templates/abstract_dataset_template.py
```diff
@@ -118,6 +118,11 @@ def coordinate_names(self) -> list[str]:
         """Returns the names of the coordinates."""
         return copy.deepcopy(self._coord_names)
 
+    @property
+    def coordinate_dimension_names(self) -> list[str]:
+        """Returns the names of the coordinate dimensions."""
+        return copy.deepcopy(self._coord_dim_names)
```
BrianMichell, line 124: A shallow copy should be sufficient for a `list[str]`, correct?

tests/unit/v1/converters/test_demultiple.py
BrianMichell, line 1: Since we don't want to bin the data this test harness should be able to be safely removed.

src/mdio/schemas/v1/templates/abstract_dataset_template.py
dmitriyrepin: Yes, the strings objects are immutable in python and self._coord_dim_names.copy() should suffice.
However, conceptually, IMHO, it is more appropriate to use copy.deepcopy() as if we were returning a list of a mutable objects, say list[Dimensions].
Fortunately, in Python, copy.deepcopy(list[str]) is identical to list[str].copy()
"String Handling:
Strings in Python are immutable. This means that once a string object is created, its value cannot be changed. Since strings are immutable, deepcopy() does not need to create new string objects for each element in the list. Instead, it places references to the same string objects from the original list into the new, deep-copied list.
Why this behavior for strings?
Because strings are immutable, there's no risk of modifying a string in one list and having that modification reflected in another list that shares the same string object. Therefore, creating new string objects would be an unnecessary overhead in terms of memory and processing. deepcopy() efficiently handles this by simply sharing references to immutable objects."

Issue #610, opened 2025-08-20
Reconcile custom SegySpecs used in the roundtrip SEGY(SegySpecs) -> MDIO(SegySpecs)-> SEGY tests


dmitriyrepin:
For the round trip tests we currently import SegY file to MDIO using a custom SegySpec
However, when export MDIO to SegY file we use the standard SegySpec
That is why the round trip test fails when it checks headers.
We need to use the same spec for both. 
How to do it is yet TBD.

tasansal:
closed by #611

PR #611, opened 2025-08-20
Export MDIO v1 to SEGY - Part 1

dmitriyrepin:
* MDIO v1 to SEGY export - PART 1  
The Part 2 will address the issue 610 "[Reconcile custom SegySpecs used in the roundtrip SEGY -> MDIO -> SEGY tests #610](https://github.com/TGSAI/mdio-python/issues/610)"
For now, header validation for roundtripping is disabled.
* Renaming of the integration tests:  
The integration tests are not independent. The test(s) of import seg_2_mdio need to be run first, then the tests that validate the structure of the MDIO, and only then the test that export mdi_2_seg can be run. The sequence of running the tests was very unclear for a newbie. The new names hopefully will make it easier for newbies to understand the order in which tests should be run. The naming also allows to properly sort the tests in the UI.
* DEBUG_MODE for tests is introduced:  
Currently all test use a temporary locations to store SEGY and MDIO files. That location changes between the test runs. After one successfully runs test(s) that import a SEG-Y to MDIO, the DEBUG_MODE allows re-running validation and export tests multiple times without re-running the import tests.

### DIFF
--- a/src/mdio/converters/mdio.py
+++ b/src/mdio/converters/mdio.py
@@ -5,12 +5,14 @@ from __future__ import annotations
 import os
 from pathlib import Path
 from tempfile import TemporaryDirectory
+from typing import TYPE_CHECKING
 
+import dask.array as da
 import numpy as np
+import xarray as xr
 from psutil import cpu_count
 from tqdm.dask import TqdmCallback
 
-from mdio import MDIOReader
 from mdio.segy.blocked_io import to_segy
 from mdio.segy.creation import concat_files
 from mdio.segy.creation import mdio_spec_to_segy
@@ -21,17 +23,33 @@ try:
 except ImportError:
     distributed = None
 
+if TYPE_CHECKING:
+    from segy.schema import SegySpec
+
+    from mdio.core.storage_location import StorageLocation
 
 default_cpus = cpu_count(logical=True)
 NUM_CPUS = int(os.getenv("MDIO__EXPORT__CPU_COUNT", default_cpus))
 
 
-def mdio_to_segy(  # noqa: PLR0912, PLR0913
-    mdio_path_or_buffer: str,
-    output_segy_path: str,
+def _get_dask_array(mdio_xr: xr.Dataset, var_name: str, chunks: tuple[int, ...] = None) -> da.Array:
+    """Workaround if the MDIO Xarray dataset returns numpy arrays instead of Dask arrays"""
+    xr_var = mdio_xr[var_name]
+    # xr_var.chunks:
+    # Tuple of block lengths for this dataarray’s data, in order of dimensions,
+    # or None if the underlying data is not a dask array.
+    if xr_var.chunks is not None:
+        return xr_var.data.rechunk(chunks)
+    # For some reason, a NumPy in-memory array was returned
+    # HACK: Convert NumPy array to a chunked Dask array
+    return da.from_array(xr_var.data, chunks=chunks)
+
+
+def mdio_to_segy(  # noqa: PLR0912, PLR0913, PLR0915
+    segy_spec: SegySpec,
+    input_location: StorageLocation,
+    output_location: StorageLocation,
     endian: str = "big",
-    access_pattern: str = "012",
-    storage_options: dict = None,
     new_chunks: tuple[int, ...] = None,
     selection_mask: np.ndarray = None,
     client: distributed.Client = None,
@@ -47,12 +65,10 @@ def mdio_to_segy(  # noqa: PLR0912, PLR0913
     A `selection_mask` can be provided (same shape as spatial grid) to export a subset.
 
     Args:
-        mdio_path_or_buffer: Input path where the MDIO is located.
-        output_segy_path: Path to the output SEG-Y file.
+        segy_spec: The SEG-Y specification to use for the conversion.
+        input_location: Store or URL (and cloud options) for MDIO file.
+        output_location: Path to the output SEG-Y file.
         endian: Endianness of the input SEG-Y. Rev.2 allows little endian. Default is 'big'.
-        access_pattern: This specificies the chunk access pattern. Underlying zarr.Array must
-            exist. Examples: '012', '01'
-        storage_options: Storage options for the cloud storage backend. Default: None (anonymous)
         new_chunks: Set manual chunksize. For development purposes only.
         selection_mask: Array that lists the subset of traces
         client: Dask client. If `None` we will use local threaded scheduler. If `auto` is used we
@@ -84,41 +100,35 @@ def mdio_to_segy(  # noqa: PLR0912, PLR0913
         ... )
 
     """
-    backend = "dask"
+    output_segy_path = Path(output_location.uri)
 
-    output_segy_path = Path(output_segy_path)
+    mdio_xr = xr.open_dataset(input_location.uri, engine="zarr", mask_and_scale=False)
 
-    mdio = MDIOReader(
-        mdio_path_or_buffer=mdio_path_or_buffer,
-        access_pattern=access_pattern,
-        storage_options=storage_options,
-    )
-
+    trace_variable_name = mdio_xr.attrs["attributes"]["traceVariableName"]
+    amplitude = mdio_xr[trace_variable_name]
+    chunks: tuple[int, ...] = amplitude.encoding.get("chunks")
+    shape: tuple[int, ...] = amplitude.shape
+    dtype = amplitude.dtype
     if new_chunks is None:
-        new_chunks = segy_export_rechunker(mdio.chunks, mdio.shape, mdio._traces.dtype)
+        new_chunks = segy_export_rechunker(chunks, shape, dtype)
+    mdio_xr.close()
 
-    creation_args = [
-        mdio_path_or_buffer,
-        output_segy_path,
-        access_pattern,
-        endian,
-        storage_options,
-        new_chunks,
-        backend,
-    ]
+    creation_args = [segy_spec, input_location, output_location, endian]
 
     if client is not None:
         if distributed is not None:
             # This is in case we work with big data
             feature = client.submit(mdio_spec_to_segy, *creation_args)
-            mdio, segy_factory = feature.result()
+            mdio_xr, segy_factory = feature.result()
         else:
             msg = "Distributed client was provided, but `distributed` is not installed"
             raise ImportError(msg)
     else:
-        mdio, segy_factory = mdio_spec_to_segy(*creation_args)
+        mdio_xr, segy_factory = mdio_spec_to_segy(*creation_args)
 
-    live_mask = mdio.live_mask.compute()
+    # Using XArray.DataArray.values should trigger compute and load the whole array into memory.
+    live_mask = mdio_xr["trace_mask"].values
+    # live_mask = mdio.live_mask.compute()
 
     if selection_mask is not None:
         live_mask = live_mask & selection_mask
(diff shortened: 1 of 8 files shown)


src/mdio/converters/mdio.py
```diff
@@ -84,41 +100,35 @@ def mdio_to_segy(  # noqa: PLR0912, PLR0913
...
-        mdio_path_or_buffer,
-        output_segy_path,
-        access_pattern,
-        endian,
-        storage_options,
-        new_chunks,
-        backend,
-    ]
+        new_chunks = segy_export_rechunker(chunks, shape, dtype)
+    mdio_xr.close()
```
BrianMichell, line 114: Why are we closing here, but using the object below still? This seems like undefined behavior.

I would recommend either using the `with` context manager or remove the `close()`

src/mdio/converters/mdio.py
```diff
@@ -84,41 +100,35 @@ def mdio_to_segy(  # noqa: PLR0912, PLR0913
...
+    output_segy_path = Path(output_location.uri)
 
-    output_segy_path = Path(output_segy_path)
-
-    mdio = MDIOReader(
-        mdio_path_or_buffer=mdio_path_or_buffer,
-        access_pattern=access_pattern,
-        storage_options=storage_options,
-    )
+    mdio_xr = xr.open_dataset(input_location.uri, engine="zarr", mask_and_scale=False)
```
BrianMichell, line 105: Maybe we can use `chunks={}` as an input parameter to ensure we use Dask arrays. The empty dict _should_ use the Zarr chunks, but this may not always be true.

[ref](https://docs.xarray.dev/en/stable/generated/xarray.open_dataset.html#xarray.open_dataset)

src/mdio/converters/mdio.py
dmitriyrepin: According to the original design (not sure why), the dataset is re-opened and returned in the call below:
```Python
mdio_xr, segy_factory = mdio_spec_to_segy(*creation_args)
```
that re-opened dataset is used later.

BrianMichell: I think that was an implementation oversight. It would be preferable to be able to pass the already opened handle to the function rather than incurring the metadata read latency multiple times. 

If @tasansal has a good reason not to do this, I would still prefer the context manager instead of closing manually.

src/mdio/converters/mdio.py
```diff
@@ -84,41 +100,35 @@ def mdio_to_segy(  # noqa: PLR0912, PLR0913
...
 
+    trace_variable_name = mdio_xr.attrs["attributes"]["traceVariableName"]
```
BrianMichell, line 107: We have a number of assumed data that may not be present. Can you please add an internal validation function that raises if the required keys are not present.

Keys:
- `mdio_xr.attrs["attributes"]` and `mdio_xr.attrs["attributes"]["traceVariableName"]`
- `mdio_xr["trace_mask"]`
- `mdio_xr["headers"]`

```python
raise UnimplementedError("This release only supports input_location files generated by `segy_to_mdio`. Required key not found.")
```

src/mdio/converters/mdio.py
```diff
@@ -138,8 +148,12 @@ def mdio_to_segy(  # noqa: PLR0912, PLR0913
...
 
     # Lazily pull the data with limits now, and limit mask so its the same shape.
-    live_mask, headers, samples = mdio[dim_slices]
-    live_mask = live_mask.rechunk(headers.chunks)
+    # Workaround: currently the MDIO Xarray dataset returns numpy arrays instead of Dask arrays
+    # TODO (Dmitriy Repin): Revisit after the eager memory allocation is fixed
+    # https://github.com/TGSAI/mdio-python/issues/608
+    live_mask = _get_dask_array(mdio_xr, "trace_mask", new_chunks[:-1])[dim_slices]
+    headers = _get_dask_array(mdio_xr, "headers", new_chunks[:-1])[dim_slices]
+    samples = _get_dask_array(mdio_xr, "amplitude", new_chunks)[dim_slices]
```
BrianMichell, line 156: We should use the key we got from the metadata.

```suggestion
    samples = _get_dask_array(mdio_xr, trace_variable_name, new_chunks)[dim_slices]
```

src/mdio/schemas/v1/templates/abstract_dataset_template.py
```diff
@@ -84,7 +84,10 @@ def build_dataset(
         self._dim_sizes = sizes
         self._horizontal_coord_unit = horizontal_coord_unit
 
-        self._builder = MDIODatasetBuilder(name=name, attributes=self._load_dataset_attributes())
+        attr = self._load_dataset_attributes() or UserAttributes(attributes={})
+        attr.attributes["traceDomain"] = self._trace_domain
```
BrianMichell, line 88: What is the use of `traceDomain` as an attributes key?

src/mdio/segy/creation.py
```diff
@@ -7,98 +7,85 @@
...
-        new_chunks=new_chunks,
-        backend=backend,
-        disk_cache=False,  # Making sure disk caching is disabled
-    )
+    mdio_xr = xr.open_dataset(input_location.uri, engine="zarr", mask_and_scale=False)
 
-    mdio_file_version = mdio.root.attrs["api_version"]
-    spec = mdio_segy_spec(mdio_file_version)
+    mdio_file_version = mdio_xr.attrs["apiVersion"]
+    spec = segy_spec
```
BrianMichell, line 70: We should just use the `segy_spec` input instead of making a copy

BrianMichell (changes requested):

src/mdio/converters/mdio.py
tasansal: +1 with Brian on this one; this would also eliminate the need for the "workaround" above.

src/mdio/converters/mdio.py
tasansal: not yet, we should use Xarray objects, no reason to re-implement any of these yet.

src/mdio/converters/mdio.py
tasansal: we don't need to close at all, just leave it, it will get replaced in the next step and object will be GC'ed.

src/mdio/schemas/v1/templates/abstract_dataset_template.py
tasansal: It is already in the template name, for consistency I prefer removing it here. more copies of something we have more inconsistencies can occur.

tests/conftest.py
tasansal, line 1: i don't like the DEBUG_MODE, please remove. this may or may not be useful for everyone's development style, this is very specific.

dmitriyrepin: During the meeting we agreed to keep the DEBUG_MODE for now and remove it during the merge to 'main'

tests/integration/test_segy_import_export_masked.py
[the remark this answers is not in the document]
dmitriyrepin: Reverted to the original test names. Added notes on the test execution order to the test docstring.

src/mdio/converters/mdio.py
dmitriyrepin: I have removed mdio_xr.close()

src/mdio/converters/mdio.py
```diff
@@ -138,8 +148,12 @@ def mdio_to_segy(  # noqa: PLR0912, PLR0913
         dim_slices += (slice(start, stop),)
 
     # Lazily pull the data with limits now, and limit mask so its the same shape.
-    live_mask, headers, samples = mdio[dim_slices]
-    live_mask = live_mask.rechunk(headers.chunks)
+    # Workaround: currently the MDIO Xarray dataset returns numpy arrays instead of Dask arrays
+    # TODO (Dmitriy Repin): Revisit after the eager memory allocation is fixed
+    # https://github.com/TGSAI/mdio-python/issues/608
+    live_mask = _get_dask_array(mdio_xr, "trace_mask", new_chunks[:-1])[dim_slices]
```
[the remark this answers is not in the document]
dmitriyrepin: Removed _get_dask_array after using Chunks={} while opening the dataset

src/mdio/segy/creation.py
```diff
@@ -7,98 +7,85 @@
...
-    mdio = MDIOReader(
-        mdio_path_or_buffer=mdio_path_or_buffer,
-        access_pattern=access_pattern,
-        storage_options=storage_options,
-        return_metadata=True,
-        new_chunks=new_chunks,
-        backend=backend,
-        disk_cache=False,  # Making sure disk caching is disabled
-    )
+    mdio_xr = xr.open_dataset(input_location.uri, engine="zarr", mask_and_scale=False)
```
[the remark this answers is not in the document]
dmitriyrepin: I have replaced all calls
```xr.open_dataset(input_location.uri, engine="zarr", mask_and_scale=False)```
with 
```open_zarr_dataset(input_location.uri)```
where
```

def open_zarr_dataset(storage_location: StorageLocation) -> xr_Dataset:
    """
    Open a Zarr dataset from the specified storage location.
    """
    # NOTE: If mask_and_scale is not set,
    # Xarray will convert int to float and replace _FillValue with NaN
    # We are using chunks={} to force Xarray to create Dask arrays
    return xr_open_dataset(storage_location.uri, engine="zarr", chunks={}, mask_and_scale=False)
```
This allows to update all xr.open_dataset parameters we require in a single location

File: src/mdio/converters/mdio.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (dmitriyrepin) [reply src/mdio/converters/mdio.py:91]: Fixed with open_zarr_dataset(..)

File: src/mdio/segy/creation.py [hunk 9]
```diff
@@ -7,98 +7,85 @@
...
-    sample_interval = sample_dim[1] - sample_dim[0]
-    samples_per_trace = len(sample_dim)
-
+    binary_header = mdio_xr.attrs["attributes"]["binaryHeader"]
+    sample_interval = binary_header["sample_interval"]
+    samples_per_trace = binary_header["samples_per_trace"]
     return SegyFactory(
         spec=spec,
-        sample_interval=sample_interval * 1000,
+        sample_interval=sample_interval,  # Sample Interval is already in microseconds
```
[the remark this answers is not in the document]
dmitriyrepin: I believe my comment is incorrect: it should say 'Sample Interval is already in **milliseconds**'.
Thus, the scalar multiplication that you request, is not needed. Sorry for misleading you.

1. The SegyFactory default sample interval is 4000. mock_nd_segy() uses the default to create the mock SegY dataset with
'sample_interval': 4000
2. segy_to_mdio stores this value (4000) in the binary_header of MDIO file.
3. mdi_to_segy retrieves the value from the binary_header["sample_interval"] as 4000
4. Now it should write it as-is to the output Seg-Y file.

If I try the change you request, I see an overflow error
```Python
key = 'sample_interval', value = 4000000
...
E OverflowError: Python integer 4000000 out of bounds for int16
```

src/mdio/converters/mdio.py
```diff
@@ -84,41 +100,35 @@ def mdio_to_segy(  # noqa: PLR0912, PLR0913
...
-    mdio = MDIOReader(
-        mdio_path_or_buffer=mdio_path_or_buffer,
-        access_pattern=access_pattern,
-        storage_options=storage_options,
-    )
+    mdio_xr = xr.open_dataset(input_location.uri, engine="zarr", mask_and_scale=False)
 
+    trace_variable_name = mdio_xr.attrs["attributes"]["traceVariableName"]
+    amplitude = mdio_xr[trace_variable_name]
+    chunks: tuple[int, ...] = amplitude.encoding.get("chunks")
```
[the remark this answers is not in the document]
dmitriyrepin: Replaced with amplitude.data.chunksize

src/mdio/converters/mdio.py
dmitriyrepin: I have implemented the change you requested

PR #616, opened 2025-08-27
Enable PreStackCocaGathers3D - take two

dmitriyrepin:
I have completely re-done the non-dimensional coordinate handling logic

### DIFF
--- a/src/mdio/converters/segy.py
+++ b/src/mdio/converters/segy.py
@@ -214,20 +214,89 @@ def populate_dim_coordinates(
     return dataset, drop_vars_delayed
 
 
-def populate_non_dim_coordinates(
+def _check_dimensions_values_identical(arr: np.ndarray, axes_to_check: tuple[int, ...]) -> np.ndarray:
+    """
+    Check if all values along specified dimensions are identical for each
+    sub-array defined by the other dimensions.
+
+    Args:
+        arr (np.ndarray): an N-dimensional array. For example, an array of all 'cdp_x' segy
+            header values for coordinates "inline", "crossline", "offset", "azimuth".
+        axes_to_check (tuple): A tuple of integers representing the axes to check for
+            identical values. For example, (2, 3) would check the "offset", "azimuth"
+            dimensions.
+
+    Returns:
+        bool: True indicates the all values in the dimensions selected by axes_to_check
+        are identical, and False otherwise.
+    """
+
+    # Create a slicing tuple to get the first element along the axes to check
+    full_slice = [slice(None)] * arr.ndim
+    for axis in axes_to_check:
+        full_slice[axis] = 0
+
+    # Broadcast the first element along the specified axes for comparison
+    first_element_slice = arr[tuple(full_slice)]
+
+    # Add new axes to the slice to enable broadcasting against the original array
+    for axis in axes_to_check:
+        first_element_slice = np.expand_dims(first_element_slice, axis)
+
+    # Compare the array with the broadcasted slice and use np.all()
+    # to collapse the dimensions being checked
+    identical = np.all(np.isclose(arr, first_element_slice), axis=axes_to_check)
+    return np.all(identical).item()
+
+
+def _populate_non_dim_coordinates(
     dataset: xr_Dataset,
     grid: Grid,
-    coordinates: dict[str, SegyHeaderArray],
+    coordinate_headers: dict[str, SegyHeaderArray],
     drop_vars_delayed: list[str],
 ) -> tuple[xr_Dataset, list[str]]:
     """Populate the xarray dataset with coordinate variables."""
-    not_null = grid.map[:] != UINT32_MAX
-    for c_name, c_values in coordinates.items():
-        c_tmp_array = dataset[c_name].values
-        c_tmp_array[not_null] = c_values
-        dataset[c_name][:] = c_tmp_array
+    coord_is_good = {}
+    # Load the grid map values into memory.
+    # Q: Should we be using the trace_mask boolean array instead of grid map?
+    grid_map_values = grid.map[:]
+    for c_name, coord_headers_values in coordinate_headers.items():
+
+        # In the case of Coca and some other templates, the coordinate header values, 
+        # coord_headers_values, have a full set of dimensions (e.g. a 4-tuple of "inline", 
+        # "crossline", "offset", "azimuth"), while the non-dimensional coordinates, (e.g., 
+        # dataset["cdp_x"]) are defined over only a subset of the dimensions (e.g. 2-tuple of 
+        # "inline", "crossline"). 
+        # Thus, every coordinate 2-tuple has multiple duplicate values of the "cdp_x" coordinates
+        # stored in coord_headers_values. Those needs to be reduced to a unique value. 
+        # We assume (and check) that all the duplicate values are (near) identical.
+        headers_dims = grid.dim_names[:-1]  # e.g.: "inline", "crossline", "offset", "azimuth"
+        coord_dims = dataset[c_name].dims   # e.g.: "inline", "crossline"
+        # This will create a temporary array in memory with the same shape as the coordinate
+        # defined in the dataset. Since the coordinate variable has not yet been populated,
+        # the temporary array will be populated with NaN from the current coordinate values.
+        tmp_nd_coord_values = dataset[c_name].values
+        # Create slices for the all grid dimensions that are also the coordinate dimensions.
+        # For other dimension, select the first element (with index 0)
+        slices = tuple(slice(None) if name in coord_dims else 0 for name in headers_dims)
+        # Create a boolean mask for the live trace values with the dimensions of the coordinate
+        # Q: Should we be using the trace_mask boolean array instead of grid map?
+        not_null = grid_map_values[slices] != UINT32_MAX
+
+        ch_reshaped = coord_headers_values.reshape(grid_map_values.shape)
+        # Select a subset of the coordinate_headers that have unique values
+        cr_reduced = ch_reshaped[slices]
+
+        # Validate the all reduced dimensions had identical values
+        axes_to_check = tuple(i for i, dim in enumerate(headers_dims) if dim not in coord_dims)
+        coord_is_good[c_name] = _check_dimensions_values_identical(ch_reshaped, axes_to_check)
+
+        # Save the unique coordinate values for live traces only
+        tmp_nd_coord_values[not_null] = cr_reduced.ravel()
+        dataset[c_name][:] = tmp_nd_coord_values
+
         drop_vars_delayed.append(c_name)
-    return dataset, drop_vars_delayed
+    return coord_is_good, dataset, drop_vars_delayed
 
 
 def _get_horizontal_coordinate_unit(segy_headers: list[Dimension]) -> AllUnits | None:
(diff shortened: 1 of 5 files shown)


src/mdio/converters/segy.py
```diff
@@ -214,20 +214,97 @@ def populate_dim_coordinates(
...
+            # Validate the all reduced dimensions had identical values
+            coord_is_good[c_name] = _check_dimensions_values_identical(ch_reshaped, axes_to_check)
+
+            # Save the unique coordinate values for live traces only
+            tmp_coord_values[not_null] = cr_reduced.ravel()
+
+        dataset[c_name][:] = tmp_coord_values
         drop_vars_delayed.append(c_name)
-    return dataset, drop_vars_delayed
+    return coord_is_good, dataset, drop_vars_delayed
```
BrianMichell, line 307: There are too many returns coming out of this function. I would prefer to exit early.

src/mdio/schemas/v1/templates/seismic_3d_prestack_coca.py
```diff
@@ -12,8 +12,8 @@ class Seismic3DPreStackCocaTemplate(AbstractDatasetTemplate):
     def __init__(self, domain: str):
         super().__init__(domain=domain)
 
-        self._coord_dim_names = ["inline", "crossline", "offset", "azimuth"]
-        self._dim_names = [*self._coord_dim_names, self._trace_domain]
+        self._coord_dim_names = ["inline", "crossline"]  # This could be a dictionary specific to every coord
```
BrianMichell, line 15: Is this comment here suggesting that we could have something like

```python
self._coord_dim_names = {"offset": ["inline", "crossline"], "azimuth": ["inline", "crossline"], "foo": ["inine", "crossline", "offset", "azimuth"]}?
```

tests/unit/v1/converters/test_segy_to_mdio.py
```diff
@@ -0,0 +1,85 @@
...
+    # NOTE: near[1][1][1][1]: np.float64(1100.0000000003308)
+
+    data_type = [
+        ("inline", "<i4"),
+        ("crossline", "<i4"),
+        ("offset", "<i4"),
+        ("azimuth", "<i4"),
+        ("cdp_x", "<f8"),
+        ("cdp_y", "<f8"),
+        ("cdp_z", "<f8"),
```
BrianMichell, line 36: I don't believe there is a formal concept of `cdp_z`. We may want to update this to something more appropriate as to not confuse people.

tests/unit/v1/converters/test_segy_to_mdio.py
BrianMichell, line 1: This file looks like it's only concerned with the convenience functions of `src/mdio/converters/segy.py`?

tests/unit/v1/converters/test_segy_to_mdio.py
```diff
@@ -0,0 +1,85 @@
+"""Tests for segY-to_mdio methods."""
```
BrianMichell, line 1:
```suggestion
"""Tests for segy_to_mdio convenience functions."""
```

src/mdio/converters/segy.py
```diff
@@ -214,20 +214,97 @@ def populate_dim_coordinates(
...
+    # Broadcast the first element along the specified axes for comparison
+    first_element_slice = arr[tuple(full_slice)]
+
+    # Add new axes to the slice to enable broadcasting against the original array
+    for axis in axes_to_check:
+        first_element_slice = np.expand_dims(first_element_slice, axis)
+
+    # Compare the array with the broadcasted slice and use np.all()
+    # to collapse the dimensions being checked
+    identical = np.all(np.isclose(arr, first_element_slice), axis=axes_to_check)
```
BrianMichell, line 248: It would make me more comfortable if we explicitly set `rtol` and `atol` to ensure consistent behavior in the event of numpy changes.

src/mdio/converters/segy.py
```diff
@@ -214,20 +214,97 @@ def populate_dim_coordinates(
...
 ) -> tuple[xr_Dataset, list[str]]:
     """Populate the xarray dataset with coordinate variables."""
-    not_null = grid.map[:] != UINT32_MAX
-    for c_name, c_values in coordinates.items():
-        c_tmp_array = dataset[c_name].values
-        c_tmp_array[not_null] = c_values
-        dataset[c_name][:] = c_tmp_array
+    coord_is_good = {}
+    # Load the grid map values into memory.
+    # Q: Should we be using the trace_mask boolean array instead of grid map?
```
BrianMichell, line 261: I believe that we want to be using the `grid.map` instead of `trace_mask` because we want to place the trace header values into the appropriate Zarr locations.

BrianMichell (changes requested):

src/mdio/converters/segy.py
```diff
@@ -214,20 +214,97 @@ def populate_dim_coordinates(
...
+            tmp_coord_values[not_null] = coord_headers_values
+        else:
+            # In the case of Coca and some other templates, the coordinate header values,
+            # coord_headers_values, have a full set of dimensions (e.g. a 4-tuple of "inline",
+            # "crossline", "offset", "azimuth"), while the non-dimensional coordinates, (e.g.,
+            # dataset["cdp_x"]) are defined over only a subset of the dimensions (e.g. 2-tuple of
+            # "inline", "crossline").
+            # Thus, every coordinate 2-tuple has multiple duplicate values of the "cdp_x" coordinates
+            # stored in coord_headers_values. Those needs to be reduced to a unique value.
+            # We assume (and check) that all the duplicate values are (near) identical.
```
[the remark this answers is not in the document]
dmitriyrepin: I am not sure. 
I suggest we re-examine this special case when we will start implementing non-binned grid overrides.

src/mdio/converters/segy.py
```diff
@@ -214,20 +214,97 @@ def populate_dim_coordinates(
...
+
+            ch_reshaped = coord_headers_values.reshape(grid_map_values.shape)
+            # Select a subset of the coordinate_headers that have unique values
+            cr_reduced = ch_reshaped[slices]
+
+            # Validate the all reduced dimensions had identical values
+            coord_is_good[c_name] = _check_dimensions_values_identical(ch_reshaped, axes_to_check)
+
+            # Save the unique coordinate values for live traces only
+            tmp_coord_values[not_null] = cr_reduced.ravel()
```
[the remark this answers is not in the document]
dmitriyrepin: Change implemented

tests/unit/v1/converters/test_segy_to_mdio.py
dmitriyrepin: To avoid confusion, I renamed
cdp_x, cdp_y, cdp_z -> cdp_diff, cdp_same, cdp_near

tests/unit/v1/converters/test_segy_to_mdio.py
dmitriyrepin: Changed

src/mdio/converters/segy.py
```diff
@@ -214,20 +214,97 @@ def populate_dim_coordinates(
...
+            # "crossline", "offset", "azimuth"), while the non-dimensional coordinates, (e.g.,
+            # dataset["cdp_x"]) are defined over only a subset of the dimensions (e.g. 2-tuple of
+            # "inline", "crossline").
+            # Thus, every coordinate 2-tuple has multiple duplicate values of the "cdp_x" coordinates
+            # stored in coord_headers_values. Those needs to be reduced to a unique value.
+            # We assume (and check) that all the duplicate values are (near) identical.
+            #
+            # The following will create a temporary array in memory with the same shape as the
+            # coordinate defined in the dataset. Since the coordinate variable has not yet been
+            # populated, the temporary array will be populated with NaN from the current coordinate
```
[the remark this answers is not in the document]
dmitriyrepin: Corrected

src/mdio/converters/segy.py
dmitriyrepin: Sorry for misunderstanding. I thought those are equivalent:
```not_null = grid_map_values[slices] != UINT32_MAX```
and
```not_null = trace_mask[slices]```

File: src/mdio/converters/segy.py [hunk 10]
```diff
@@ -214,20 +214,97 @@ def populate_dim_coordinates(
...
+            # values.
+            tmp_coord_values = dataset[c_name].values
```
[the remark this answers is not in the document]
dmitriyrepin: The PR [Eager memory allocation fix #609](https://github.com/TGSAI/mdio-python/pull/609/files#top) replaced
```dataset[c_name].values[not_null] = c_values```
with
```

c_tmp_array = dataset[c_name].values
c_tmp_array[not_null] = c_values
dataset[c_name][:] = c_tmp_array
```
My understanding of the change is that it fixed non-updating dataset[c_name].
 "You cannot modify a Dask array in place. Dask arrays are designed to be immutable, meaning that all operations, including what looks like an in-place modification, will return a new Dask array with an updated task graph". Thus, creating a modified copy of the dask array c_tmp_array is unavoidable.
Please correct me, if needed.

File: src/mdio/converters/segy.py [hunk 11]
```diff
@@ -214,20 +214,97 @@ def populate_dim_coordinates(
...
     """Populate the xarray dataset with coordinate variables."""
-    not_null = grid.map[:] != UINT32_MAX
-    for c_name, c_values in coordinates.items():
-        c_tmp_array = dataset[c_name].values
-        c_tmp_array[not_null] = c_values
-        dataset[c_name][:] = c_tmp_array
+    coord_is_good = {}
+    # Load the grid map values into memory.
+    # Q: Should we be using the trace_mask boolean array instead of grid map?
+    grid_map_values = grid.map[:]
```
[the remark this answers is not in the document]
dmitriyrepin: The original code had
```not_null = grid.map**[:]** != UINT32_MAX```

Thus, am not sure the following would lead to the expected result.
```Python
if axes_to_check == ():
   not_null = grid.map.data != UINT32_MAX
   ...
else:
   not_null = grid.map[slices].data != UINT32_MAX```
   ...
```
Please, advise.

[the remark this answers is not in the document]
BrianMichell: I believe the appropriate way to manage the lifetime of the materialized array would be:

```python
if axes_to_check == ():
    not_null = grid.map[:] != UINT32_MAX
   ...
else:
    not_null = grid.map[slices] != UINT32_MAX
    ...
```

This could also be a ternary assignment, however I am opposed to using those in almost all scenarios.

src/mdio/converters/segy.py
```diff
@@ -214,18 +214,89 @@ def populate_dim_coordinates(
...
-        c_tmp_array[not_null] = c_values
-        dataset[c_name][:] = c_tmp_array
+    # Load the grid map values into memory.
+    for c_name, coord_headers_values in coordinate_headers.items():
+        headers_dims = grid.dim_names[:-1]  # e.g.: "inline", "crossline", "offset", "azimuth"
+        coord_dims = dataset[c_name].dims  # e.g.: "inline", "crossline"
+        axes_to_check = tuple(i for i, dim in enumerate(headers_dims) if dim not in coord_dims)
+        if axes_to_check == ():
+            # In case the coordinate has the same dimensions as grid map
+            not_null = grid.map[:] != UINT32_MAX
```
BrianMichell, line 266:
```suggestion
            not_null = grid.map[:] != grid.map.fill_value
```

Missed this in previous reviews. The type and fill value of `grid.map` may be internally promoted to `uint64`.

src/mdio/converters/segy.py
```diff
@@ -214,18 +214,89 @@ def populate_dim_coordinates(
...
+            # The following will create a temporary array in memory with the same shape as the
+            # coordinate defined in the dataset. Since the coordinate variable has not yet been
+            # populated, the temporary array will be populated with _FillValue from the current
+            # coordinate values.
+            tmp_coord_values = dataset[c_name].values
+            # Create slices for the all grid dimensions that are also the coordinate dimensions.
+            # For other dimension, select the first element (with index 0)
+            slices = tuple(slice(None) if name in coord_dims else 0 for name in headers_dims)
+            # Create a boolean mask for the live trace values with the dimensions of the coordinate
+            not_null = grid.map[slices] != UINT32_MAX
```
BrianMichell, line 288:
```suggestion
            not_null = grid.map[slices] != grid.map.fill_value
```

Same here

BrianMichell (changes requested):

BrianMichell (approved):
LGTM
@tasansal

tasansal (changes requested):
this implementation is not valid. as i said before the coordinates don't always share the same dimensions we have to handle this more explicitly.

PR #620, opened 2025-09-02
On-the-fly ingestion chunking

### LINKED ISSUE #587: Enhance segy_to_mdio_v1 to do chunked write for non-dimensional coordinates and trace_mask
Currently we write to Zarr in two steps. 
1) We write the xarray dataset to Zarr with to create the Zarr store with the correct structure as following:
    - Populated arrays:
      - 1D dimensional coordinates
      - ND non-dimensional coordinates
      - ND trace_mask 
    - Empty arrays (will be populated later in chunks):
      - ND+1 traces
      - ND headers (no _FillValue set due to the bug https://github.com/TGSAI/mdio-python/issues/582)
2) Write the headers and traces in chunks using grid_map to indicate dead traces

It has been suggested that for extremely large datasets we could  do chunked write for non-dimensional coordinates and trace_mask

BrianMichell:
Resolves #587 

While this does resolve the issue, it's not clean and doesn't happen at the stage I would like. There are also no tests to ensure appropriate behavior of dataset mutation.

### DIFF
--- a/src/mdio/converters/segy.py
+++ b/src/mdio/converters/segy.py
@@ -18,10 +18,19 @@ from mdio.converters.exceptions import GridTraceCountError
 from mdio.converters.exceptions import GridTraceSparsityError
 from mdio.converters.type_converter import to_structured_type
 from mdio.core.grid import Grid
+from mdio.core.utils_write import MAX_COORDINATES_BYTES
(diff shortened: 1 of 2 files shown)


tasansal:
this doesn't look bad, I'll rebase and make some tweaks and merge tomorrow.

PR #629, opened 2025-09-04
Add docs for installing with uv.

markspec:


tasansal:
The usage of `uv` is already documented in [CONTRIBUTING.md](https://github.com/TGSAI/mdio-python/blob/main/CONTRIBUTING.md#how-to-set-up-your-development-environment)

It is not needed for user documentation. The only packaging that is relevant to MDIO users are `pip` and `conda`. `uv` encapsulates `pip` and `virtualenv`, and a developer or user's choice to use it or not.

PR #659, opened 2025-09-10
Environment variable flag to preserve unmodified, raw trace headers from SEG-Y as variable.

BrianMichell:
Supersedes #634 

- Only supports Zarr v3 configuration since v2 handles the raw bytes differently.
- Ensures we retain endianness of the raw headers.

Pending work
- [x] Memory and CPU profiling

Notes:
Requires environment variable `MDIO__DO_RAW_HEADERS` `1`
Will generate a logger warning regarding experimental status if the feature is enabled.


tests/integration/test_segy_import_export_masked.py
tasansal, line 1: i see no tests here? ARG002 also suggests that variable isn't used

src/mdio/segy/_workers.py
```diff
@@ -121,23 +122,31 @@ def trace_worker(  # noqa: PLR0913
...
     # Get subset of the dataset that has not yet been saved
     # The headers might not be present in the dataset
     worker_variables = [data_variable_name]
     if header_key in dataset.data_vars:  # Keeping the `if` here to allow for more worker configurations
         worker_variables.append(header_key)
+    if raw_header_key in dataset.data_vars:
+        do_reverse_transforms = True
+        worker_variables.append(raw_header_key)
 
+    raw_headers, transformed_headers, traces = get_header_raw_and_transformed(
```
tasansal, line 141:
```suggestion
    header_pipeline = deepcopy(segy_file.accessors.header_decode_pipeline)
    segy_file.accessors.header_decode_pipeline.transforms = []
```

Reversing transforms is not needed. you can extract the header transform pipeline and defer its application later. The above snipped I posted will skip all transforms, and you can write that out as `raw_headers`. Then `header_pipeline.apply(...)` will generate the interpreted versions and you can write them out as `headers. This way the `_disaster_recovery_wrapper.py` and all its tests can also be removed.  would still recommend keeping the tests that use a mock segy file of course.

tasansal (changes requested):

tests/integration/test_segy_import_export_masked.py
BrianMichell: I added the `raw_headers_env` fixture as a config option for the ingestion tests.

Had to be passed to the functions but not used.

tasansal (approved):
LGTM but hopefully we can remove this soon!

PR #660, opened 2025-09-10
Re-implement on-the-fly chunking for non-dimension coordinates and li…

BrianMichell:
Supersedes #620 

Same logic is implemented with only changes being how the Pydantic models are used.


tasansal (approved):
LGTM, we can tweak this later if needed.

PR #666, opened 2025-09-12
Fix for hardcoded fill value for local grid map in trace worker

### LINKED ISSUE #661: Indexing sentinel logic issue
https://github.com/TGSAI/mdio-python/blob/4cea31bb2735f75de8cc738e5780b2196512568e/src/mdio/segy/_workers.py#L110

We need to ensure that we use the proper sentinel. This could be promoted to UINT64_MAX on larger grids.

May have similar bugs elsewhere.

BrianMichell:
Resolves #661


tasansal:
LGTM, but what is puzzling to me is, how did the patch coverage go to 50% ?

tasansal (approved):

PR #673, opened 2025-09-15
Ensure no orphaned chunks on Dataset overwrite

### LINKED ISSUE #656: SEG-Y Ingestion overwrite bug
It appears that in `segy_to_mdio` with `overwrite=True` if a different schema is provided the following error is produced.

```bash
ValueError: dataset contains non-pre-existing variables ['cdp_x'], which is not allowed in ``xarray.Dataset.to_zarr()`` with ``mode='r+'``. To allow writing new variables, set ``mode='a'``.
```

Original report stated that the original file had an MDIO 0.9 dataset present when the error was generated.

We should either ensure that overwrite will not leave any orphaned data, warn that orphaned data may occur, or fail if input and existing schemas are incompatible.

dmitriyrepin:
Resolves #656


BrianMichell (changes requested):
I do not like this implementation, it runs the risk of deleting things that are not MDIO files. I would instead ensure that the API is able to open the file before any deletion operation is solicited. See the [C++ implementation](https://github.com/TGSAI/mdio-cpp/blob/main/mdio/utils/delete.h).

Issue #613, opened 2025-08-26
Open handles not properly cleaned up


BrianMichell:
When ingesting `segy_to_mdio` in cloud-to-cloud mode file handles do not get properly cleaned up. This leads to many errors being printed to stdout.

These are non-fatal, but can be very misleading.

```bash
Scanning SEG-Y for geometry attributes: 100%|██████████| 3013/3013 [02:00<00:00, 24.96block/s]
Ingesting traces: 100%|██████████| 2303/2303 [05:43<00:00,  6.70block/s] 
Traceback (most recent call last):
Traceback (most recent call last):
Traceback (most recent call last):
Traceback (most recent call last):
Traceback (most recent call last):
Traceback (most recent call last):
Traceback (most recent call last):
Traceback (most recent call last):
  File "uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/weakref.py", line 666, in _exitfunc
    f()
    ~^^
  File "uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/weakref.py", line 590, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
           ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.13/site-packages/gcsfs/core.py", line 380, in close_session
    asyn.sync(asyn.loop[0], session.close, timeout=0.1)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.13/site-packages/fsspec/asyn.py", line 103, in sync
    raise return_result
  File ".venv/lib/python3.13/site-packages/fsspec/asyn.py", line 56, in _runner
    result[0] = await coro
                ^^^^^^^^^^
  File "uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/asyncio/tasks.py", line 507, in wait_for
    return await fut
           ^^^^^^^^^
  File "uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/weakref.py", line 666, in _exitfunc
    f()
    ~^^
  File ".venv/lib/python3.13/site-packages/aiohttp/client.py", line 1327, in close
    await self._connector.close()
  File "uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/weakref.py", line 590, in __call__
[...]

PR #680, opened 2025-09-22
Fix bug 613 Open handles not properly cleaned up v1

dmitriyrepin:
This is an ugly fix for the bug 613  [Open handles not properly cleaned up](https://github.com/TGSAI/mdio-python/issues/613)
The bug is caused by the [Zarr issue #3487](https://github.com/zarr-developers/zarr-python/issues/3487). 
As a workaround for the issue:
1. remove any direct or indirect (e.g., via SegyFile) references to fsspec from the main thread
2. Reading of SegyFile is done in separate processes


dmitriyrepin:
This PR is closed and changes are transferred to PR https://github.com/TGSAI/mdio-python/pull/701

Issue #681, opened 2025-09-23
Disaster recovery of binary header


BrianMichell:
We must preserve the raw 400 bytes of the SEG-Y Binary header in MDIO. This should probably be an extension of the [_add_segy_file_headers](https://github.com/TGSAI/mdio-python/blob/320b392268f7abae97eaa6c208f523000bd6d6ea/src/mdio/converters/segy.py#L296) function.

To keep the type inference pipeline from running please see the [suggestion](https://github.com/TGSAI/mdio-python/pull/659#discussion_r2362608026) from @tasansal. This PR may provide other useful information.

Additional information helpful for this task:
- Consider a structure like `np.ascontiguousarray` to ensure we don't copy garbage values.
- Using the `.view("|V400")` should ensure we get all bytes.
- If it proves difficult to ensure full preservation of data without modifying the original outputs from `_add_segy_file_headers` we can consider re-opening the `SegyFile`.

Success criteria:
- Integration test added to [import/export](https://github.com/TGSAI/mdio-python/blob/main/tests/integration/test_segy_import_export_masked.py) which compares the segy bytes on disk to bytes in MDIO.
```python
def test_raw_binary_header_preservation(
    self, test_conf, export_masked_path):
```
- Ensure this works for both standard and non-standard binary headers. Use best judgement for where to update/add test configuration. Faked headers such as "foo" and "bar" are acceptable. [Rev 1 standard](https://library.seg.org/pb-assets/technical-standards/seg_y_rev1-1686080991247.pdf) (page 9).

PR #682, opened 2025-09-24
Binary header disaster recovery

dmitriyrepin:
Addresses the issue [#681 Disaster recovery of binary header ](https://github.com/TGSAI/mdio-python/issues/681)
## Requires:
* PR #680

## Approach
SegyFile currently reads and processes the raw binary headers in its cached property `binary_header`.
The logic of that property can be divided in two steps: first, reading the raw binary header and second, converting it to HeaderArray
Then mdio can 
1. Directly read the raw binary header and use it for disaster recovery
2. Second, convert the raw binary header HeaderArray using logic from the SEG-Y library to store in the MDIO headers

If the proposed approach is acceptable, some of the PR code should be moved to the SEG-Y library.

## NeedMoreInfo
Where should we save the raw binary header for the disaster recovery?


BrianMichell:
Please see [fallback](https://github.com/BrianMichell/mdio-python/pull/9/files#diff-30d66960dcbe7846b677f6f93f10c41cbd06fafb9079aa99098c23d2536dca26) implementation for similar trace header DR array. Note that due to an undetermined bug the file was customized such that any unspecified byte was added as a uint8.

My recommendation would be to update your implementation to follow this pattern and drop the unspecified fields from the `binary_header_dict`.

We should also add a new Variable `raw_binary_header` of type `r3200` and rank 1. This is where the DR array should be written to. It can be serialized at the same time as the coordinates and dimensions.

PR #683, opened 2025-09-25
Environment variable flag to preserve unmodified, raw binary header from SEG-Y in `segy_file_header` variable.

dmitriyrepin:
## Addresses 
issue #681  issue Disaster recovery of binary header

## Approach
SegyFile currently reads and processes the raw binary headers in its cached property binary_header.
The logic of that property can be divided in two steps: first, reading the raw binary header and second, converting it to HeaderArray

Then mdio can

1. Directly read the raw binary header and use it for disaster recovery
2. Second, convert the raw binary header HeaderArray using logic from the SEG-Y library to store in the MDIO headers

If the proposed approach is acceptable, some of the PR code should be moved to the SEG-Y library.

### DIFF
--- a/tests/integration/test_segy_roundtrip_teapot.py
+++ b/tests/integration/test_segy_roundtrip_teapot.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import json
-import os
 from typing import TYPE_CHECKING
 
 import dask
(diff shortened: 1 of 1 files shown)


BrianMichell (approved):
Not the final version, but seems fundamentally solid to me

PR #685, opened 2025-09-26
clean tests and make sure all test code is executed

dmitriyrepin:
Clean tests and make sure all test code is executed


tests/unit/conftest.py
```diff
@@ -1,59 +0,0 @@
-"""Extra configurations for unit tests."""
```
BrianMichell, line 1: Are we positive that these configurations are fine to go away? I believe that we may want to retain this for some of the deferred features.

BrianMichell (changes requested):
I believe that this PR is the appropriate place to implement the removal of the debug mode.

dmitriyrepin:
> I believe that this PR is the appropriate place to implement the removal of the debug mode.
DEBUG_MODE has been removed

PR #687, opened 2025-09-29
create empty

dmitriyrepin:
Add new API
```Python
def create_empty_mdio(  # noqa PLR0913
    segy_spec: SegySpec,
    mdio_template: AbstractDatasetTemplate,
    grid: Grid,
    output_path: UPath | Path | str,
    overwrite: bool = False,
) -> None:
    """A function that creates an empty MDIO v1 file with known dimensions.

    Args:
        segy_spec: The SEG-Y specification to use for trace headers.
        mdio_template: The MDIO template to use to define the dataset structure.
        grid: The grid specifying the dimensions of the MDIO file.
        output_path: The universal path for the output MDIO v1 file.
        overwrite: Whether to overwrite the output file if it already exists. Defaults to False.
```

### DIFF
--- a/src/mdio/converters/segy.py
+++ b/src/mdio/converters/segy.py
@@ -417,3 +417,61 @@ def segy_to_mdio(  # noqa PLR0913
         dataset=xr_dataset,
         data_variable_name=default_variable_name,
     )
+
+
+def create_empty_mdio(  # noqa PLR0913
+    segy_spec: SegySpec,
+    mdio_template: AbstractDatasetTemplate,
+    grid: Grid,
+    output_path: UPath | Path | str,
+    overwrite: bool = False,
+) -> None:
+    """A function that creates an empty MDIO v1 file with known dimensions.
+
+    Args:
+        segy_spec: The SEG-Y specification to use for trace headers.
+        mdio_template: The MDIO template to use to define the dataset structure.
+        grid: The grid specifying the dimensions of the MDIO file.
+        output_path: The universal path for the output MDIO v1 file.
+        overwrite: Whether to overwrite the output file if it already exists. Defaults to False.
+
+    Raises:
+        FileExistsError: If the output location already exists and overwrite is False.
+    """
+    output_path = _normalize_path(output_path)
+
+    if not overwrite and output_path.exists():
+        err = f"Output location '{output_path.as_posix()}' exists. Set `overwrite=True` if intended."
+        raise FileExistsError(err)
+
+    # Build the dataset structure using the template and grid
+    header_dtype = to_structured_type(segy_spec.trace.header.dtype)
+    horizontal_unit = _get_horizontal_coordinate_unit(grid.dims)
+    mdio_ds: Dataset = mdio_template.build_dataset(
+        name=mdio_template.name,
+        sizes=grid.shape,
+        horizontal_coord_unit=horizontal_unit,
+        header_dtype=header_dtype,
+    )
+
+    # Convert to xarray dataset
+    xr_dataset: xr_Dataset = to_xarray_dataset(mdio_ds=mdio_ds)
+
+    # Populate coordinates using the grid
+    # For empty datasets, we only populate dimension coordinates
+    drop_vars_delayed = []
+    dataset, drop_vars_delayed = populate_dim_coordinates(xr_dataset, grid, drop_vars_delayed=drop_vars_delayed)
+
+    # Set the trace mask to indicate all traces are live (since this is an empty dataset)
+    if grid.live_mask is not None:
+        dataset.trace_mask.data[:] = grid.live_mask
+    else:
+        # If live_mask is None, create a mask where all traces are live
+        dataset.trace_mask.data[:] = True
+
+    # Create the Zarr store with the correct structure but with empty arrays
+    to_mdio(dataset, output_path=output_path, mode="w", compute=False)
+
+    # Write the dimension coordinates and trace mask
+    meta_ds = dataset[drop_vars_delayed + ["trace_mask"]]
+    to_mdio(meta_ds, output_path=output_path, mode="r+", compute=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,7 +23,7 @@ warnings.filterwarnings(
 def fake_segy_tmp(tmp_path_factory: pytest.TempPathFactory) -> Path:
     """Make a temp file for the fake SEG-Y files we are going to create."""
     if DEBUG_MODE:
-        return Path("TMP/fake_segy")
+        return Path("tmp/fake_segy")
     return tmp_path_factory.mktemp(r"fake_segy")
 
 
@@ -37,7 +37,7 @@ def segy_input_uri() -> str:
 def segy_input(segy_input_uri: str, tmp_path_factory: pytest.TempPathFactory) -> Path:
     """Download teapot dome dataset for testing."""
     if DEBUG_MODE:
-        tmp_dir = Path("TMP/segy")
+        tmp_dir = Path("tmp/segy")
         tmp_dir.mkdir(parents=True, exist_ok=True)
     else:
         tmp_dir = tmp_path_factory.mktemp("segy")
@@ -50,7 +50,7 @@ def segy_input(segy_input_uri: str, tmp_path_factory: pytest.TempPathFactory) ->
 def zarr_tmp(tmp_path_factory: pytest.TempPathFactory) -> Path:
     """Make a temp file for the output MDIO."""
     if DEBUG_MODE:
-        return Path("TMP/mdio")
+        return Path("tmp/mdio")
     return tmp_path_factory.mktemp(r"mdio")
 
 
@@ -58,7 +58,7 @@ def zarr_tmp(tmp_path_factory: pytest.TempPathFactory) -> Path:
 def zarr_tmp2(tmp_path_factory: pytest.TempPathFactory) -> Path:
     """Make a temp file for the output MDIO."""
     if DEBUG_MODE:
-        return Path("TMP/mdio2")
+        return Path("tmp/mdio2")
     return tmp_path_factory.mktemp(r"mdio2")
 
 
@@ -66,8 +66,16 @@ def zarr_tmp2(tmp_path_factory: pytest.TempPathFactory) -> Path:
 def segy_export_tmp(tmp_path_factory: pytest.TempPathFactory) -> Path:
     """Make a temp file for the round-trip IBM SEG-Y."""
     if DEBUG_MODE:
-        tmp_dir = Path("TMP/segy")
+        tmp_dir = Path("tmp/segy")
         tmp_dir.mkdir(parents=True, exist_ok=True)
     else:
         tmp_dir = tmp_path_factory.mktemp("segy")
     return tmp_dir / "teapot_roundtrip.segy"
+
+
+@pytest.fixture(scope="class")
+def empty_mdio(tmp_path_factory: pytest.TempPathFactory) -> Path:
+    """Make a temp file for empty MDIO testing."""
+    if DEBUG_MODE:
+        return Path("tmp/empty_mdio")
+    return tmp_path_factory.mktemp(r"empty_mdio")
--- a/tests/integration/test_create_empty_mdio.py
+++ b/tests/integration/test_create_empty_mdio.py
@@ -1,0 +1,122 @@
+"""Test for create_empty_mdio function."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import numpy as np
+import pytest
+from segy.standards import get_segy_standard
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from segy.schema import SegySpec
+from tests.integration.testing_helpers import get_values
+from tests.integration.testing_helpers import validate_variable
+
+from mdio import __version__
+from mdio.api.io import open_mdio
+from mdio.builder.template_registry import get_template
+from mdio.converters.segy import create_empty_mdio
+from mdio.core import Dimension
+from mdio.core import Grid
+
+
+class TestCreateEmptyPostStack3DTimeMdio:
+    """Tests for create_empty_mdio function."""
+
+    @pytest.fixture(scope="class")
+    def segy_spec(self) -> SegySpec:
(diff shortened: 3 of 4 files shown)


tasansal:
Thanks Dmitry. We should make it less specific to SEGY. I propose the following design:

1. Users need to use or register a new template
2. `create_empty` takes just the template and grid, no segy_spec.
3. move this out of converters, have its own creation module


Additionally, the tutorial notebook needs to be updated

BrianMichell:
To add some further context

1. Headers and live mask are not mandatory for MDIO V1. This is a change that should be noted in the documentation. Additionally, since these Variables are not required we should notate this in the [Write to SEG-Y](https://mdio-python.readthedocs.io/en/v0.9.3/tutorials/creation.html#write-to-seg-y) section.
2. The [copy_empty_like](https://mdio-python.readthedocs.io/en/v0.9.3/tutorials/creation.html#write-to-seg-y) has not been deeply discussed. Please create an issue for this feature and link it in the documentation. We will discuss the appropriate implementation but feel free to propose the appropriate behavior.

src/mdio/converters/segy.py
```diff
@@ -291,7 +291,7 @@ def populate_non_dim_coordinates(
     return dataset, drop_vars_delayed
 
 
-def _get_horizontal_coordinate_unit(segy_headers: list[Dimension]) -> LengthUnitModel | None:
+def get_horizontal_coordinate_unit(segy_headers: list[Dimension]) -> LengthUnitModel | None:
```
BrianMichell, line 294:
```suggestion
def _get_horizontal_coordinate_unit(segy_headers: list[Dimension]) -> LengthUnitModel | None:
```

I don't think we want to denote this as a public function.

src/mdio/creators/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""MDIO Data creation API."""
+
+from mdio.creators.mdio import create_empty_mdio
```
BrianMichell, line 3:
```suggestion
from mdio.creation.mdio import create_empty_mdio
```

Thoughts on `creation` over `creators` for module name?

src/mdio/creators/mdio.py
```diff
@@ -0,0 +1,80 @@
...
+    from pathlib import Path
+
+    from upath import UPath
+    from xarray import Dataset as xr_Dataset
+
+    from mdio.builder.schemas import Dataset
+    from mdio.core.dimension import Dimension
+
+
+def create_empty_mdio(  # noqa PLR0913
```
BrianMichell, line 28:
```suggestion
def create_empty(  # noqa PLR0913
```

Suffixing everything as `_mdio` feels redundant to me

src/mdio/creators/mdio.py
```diff
@@ -0,0 +1,80 @@
...
+    mdio_template_name: str,
+    dimensions: list[Dimension],
+    output_path: UPath | Path | str,
+    create_headers: bool = False,
```
BrianMichell, line 32:
```suggestion
```

I don't think we should include this option at all. If the user wants to create headers it should be part of their template. We want to get away from SEG-Y as much as possible.

src/mdio/creators/mdio.py
```diff
@@ -0,0 +1,80 @@
...
+    # Convert to xarray dataset
+    xr_dataset: xr_Dataset = to_xarray_dataset(mdio_ds=mdio_ds)
+
+    # Populate coordinates using the grid
+    # For empty datasets, we only populate dimension coordinates
+    drop_vars_delayed = []
+    dataset, drop_vars_delayed = populate_dim_coordinates(xr_dataset, grid, drop_vars_delayed=drop_vars_delayed)
+
+    # Set the trace mask to indicate all traces are live (since this is an empty dataset)
+    dataset.trace_mask.data[:] = True
```
BrianMichell, line 73:
```suggestion
```

Since we are creating an empty dataset only essential data should be written. To me that means ensuring `trace_mask` is set to the fill value so that we don't serialize it to disk.

src/mdio/creators/mdio.py
```diff
@@ -0,0 +1,80 @@
...
+
+    if not overwrite and output_path.exists():
+        err = f"Output location '{output_path.as_posix()}' exists. Set `overwrite=True` if intended."
+        raise FileExistsError(err)
+
+    header_dtype = to_structured_type(get_segy_standard(1.0).trace.header.dtype) if create_headers else None
+    grid = Grid(dims=dimensions)
+    horizontal_unit = get_horizontal_coordinate_unit(grid.dims)
+    mdio_template = TemplateRegistry().get(mdio_template_name)
+    mdio_ds: Dataset = mdio_template.build_dataset(
```
BrianMichell, line 57: Referring to my comments on line 32 and 73, should the `build_dataset` method always be calling the `_add_trace_mask` method? I'm not certain that this should be default behavior coming out of the `AbstractDatasetTemplate`. Thoughts @tasansal @dmitriyrepin ?

src/mdio/creators/mdio.py
```diff
@@ -0,0 +1,80 @@
...
+        output_path: The universal path for the output MDIO v1 file.
+        create_headers: Whether to create a full set of SEG-Y v1.0 trace headers. Defaults to False.
+        overwrite: Whether to overwrite the output file if it already exists. Defaults to False.
+
+    Raises:
+        FileExistsError: If the output location already exists and overwrite is False.
+    """
+    output_path = _normalize_path(output_path)
+
+    if not overwrite and output_path.exists():
```
BrianMichell, line 49: Lines 47-51 now exist in more than one place. It may make sense to refactor this into a convenience function under `mdio.api`? Then we can repurpose test `test_overwrite_behavior` for the appropriate file exists message raised and trust the other libraries to behave as expected in respect to overwrite.

BrianMichell (changes requested):

src/mdio/creators/mdio.py
dmitriyrepin: Removed

src/mdio/creators/mdio.py
dmitriyrepin: As suggested by @tasansal , replaced
```create_headers: bool = False,```
with
```headers: HeaderSpec | None = None,```

File: src/mdio/creators/mdio.py (hunk 6 above)
Comment (dmitriyrepin) [reply src/mdio/creators/mdio.py:54]: We will updated create_empty() when the default behavior of `AbstractDatasetTemplate` changes.
For now, when the 'trace_mask' variable is created, it contains the fill values ( 'False' - all traces are marked as dead)

File: src/mdio/api/create.py [hunk 8]
```diff
@@ -0,0 +1,175 @@
...
+    ds = open_mdio(input_path)
+
+    # Create a copy with the same structure but no data or,
+    # optionally, coordinates
+    ds_output = ds.copy(data=None).reset_coords(drop=not keep_coordinates)
+
+    # Dataset
+    # Keep the name (which is the same as the used template name) and the original API version
+    # ds_output.attrs["name"]
+    # ds_output.attrs["apiVersion"]
```
BrianMichell: Should we retain the original apiVersion or the apiVersion used with `createEmptyLike`?

src/mdio/api/create.py
```diff
@@ -0,0 +1,175 @@
...
+        # attr["defaultVariableName"]
+        # attr["surveyType"]
+        # attr["gatherType"]
+
+    # "All traces should be marked as dead in empty dataset"
+    if "trace_mask" in ds_output.variables:
+        ds_output["trace_mask"][:] = False
+
+    # Data variable
+    var_name = attr["defaultVariableName"]
```
BrianMichell: `defaultVariableName` is not a required field. Please validate that it is present before getting the attribute.

`statsV1` is also not a required field for a given `defaultVariableName`.

tests/integration/test_segy_roundtrip_teapot.py
```diff
@@ -156,25 +151,29 @@ class TestTeapotRoundtrip:
...
-        zarr_tmp: Path,
+        teapot_mdio_tmp: Path,
         teapot_segy_spec: SegySpec,
     ) -> None:
         """Test importing a SEG-Y file to MDIO.
 
         NOTE: This test must be executed before the 'TestReader' and 'TestExport' tests.
         """
+        unit_aware_template = TemplateRegistry().get("PostStack3DTime")
+        unit_aware_template.add_units({"time": UNITS_SECOND})
```
BrianMichell: Units should be `ms`

tests/integration/test_z_create_empty.py
```diff
@@ -0,0 +1,408 @@
...
+
+    @classmethod
+    def _create_empty_mdio(cls, create_headers: bool, output_path: Path, overwrite: bool = True) -> xr_Dataset:
+        """Create a temporary empty MDIO file for testing."""
+        # Create the grid with the specified dimensions
+        dims = [
+            Dimension(name="inline", coords=range(1, 346, 1)),  # 100-300 with step 1
+            Dimension(name="crossline", coords=range(1, 189, 1)),  # 1000-1600 with step 2
+            Dimension(name="time", coords=range(0, 3002, 2)),  # 0-3 seconds 4ms sample rate
+        ]
```
BrianMichell: I believe the comments here are incorrect

tests/integration/test_z_create_empty.py
```diff
@@ -0,0 +1,408 @@
...
+        self.validate_teapod_dataset_metadata(ds, is_velocity=True)
+
+    def test_variables(self, mdio_with_headers: Path, mdio_no_headers: Path) -> None:
+        """Test grid validation for empty MDIO file."""
+        ds = open_mdio(mdio_with_headers)
+        header_dtype = get_teapot_segy_spec().trace.header.dtype
+        self.validate_teapod_dataset_variables(ds, header_dtype=header_dtype, is_velocity=True)
+
+        ds = open_mdio(mdio_no_headers)
+        self.validate_teapod_dataset_variables(ds, header_dtype=None, is_velocity=True)
```
BrianMichell:
```suggestion
        self.validate_teapot_dataset_variables(ds, header_dtype=None, is_velocity=True)
```

tests/integration/test_z_create_empty.py
```diff
@@ -0,0 +1,408 @@
...
+        with pytest.raises(FileExistsError, match="Output location.*exists"):
+            self._create_empty_mdio(create_headers=True, output_path=empty_mdio, overwrite=False)
+
+        # Third call: Create MDIO with overwrite=True - should succeed and overwrite garbage
+        xr_dataset = self._create_empty_mdio(create_headers=True, output_path=empty_mdio, overwrite=True)
+        assert xr_dataset is not None
+
+        # Validate that the MDIO file can be loaded correctly using the helper function
+        ds = open_mdio(empty_mdio)
+        self.validate_teapod_dataset_metadata(ds, is_velocity=True)
```
BrianMichell:
```suggestion
        self.validate_teapot_dataset_metadata(ds, is_velocity=True)
```

BrianMichell (changes requested):
`teapod` should be fixed to `teapot`. I think I caught all instance but please double check.

It looks like there was some scope creep in this PR which started updating tests like the `test_segy_roundtrip_teapot` source file. If you could please remove these and open a separate PR to keep changes limited in scope and reviews managable.

src/mdio/api/create.py
```diff
@@ -0,0 +1,175 @@
...
+        xr_dataset["segy_file_header"] = ((), "")
+
+    # Create the Zarr store with the correct structure but with empty arrays
+    if output_path is not None:
+        to_mdio(xr_dataset, output_path=output_path, mode="w", compute=False)
+
+    # Write the dimension coordinates and trace mask
+    xr_dataset = xr_dataset[drop_vars_delayed + ["trace_mask"]]
+
+    if output_path is not None:
```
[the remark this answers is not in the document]
dmitriyrepin: The `None` output path allows to create in-memory only MDIO dataset (if the data size is small enough) to be used as a transient storage. One of such uses would be running in-browser code examples by the users that do not have access to the cloud storage.

tests/integration/test_z_create_empty.py
```diff
@@ -0,0 +1,408 @@
...
+    def test_dataset_metadata(self, mdio_with_headers: Path) -> None:
+        """Test dataset metadata for empty MDIO file."""
+        ds = open_mdio(mdio_with_headers)
+        self.validate_teapod_dataset_metadata(ds, is_velocity=True)
+
+    def test_variables(self, mdio_with_headers: Path, mdio_no_headers: Path) -> None:
+        """Test grid validation for empty MDIO file."""
+        ds = open_mdio(mdio_with_headers)
+        header_dtype = get_teapot_segy_spec().trace.header.dtype
+        self.validate_teapod_dataset_variables(ds, header_dtype=header_dtype, is_velocity=True)
```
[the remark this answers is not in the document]
dmitriyrepin: Corrected. No "teapod" string(s) are present on the codebase anymore.

tests/integration/test_z_create_empty.py
dmitriyrepin: Corrected. No "teapod" string(s) are present on the codebase anymore.

tests/integration/test_z_create_empty.py
dmitriyrepin: Corrected. No "teapod" string(s) are present on the codebase anymore.

tests/integration/test_z_create_empty.py
dmitriyrepin: You are absolutely correct. The comments are removed. 
The were leftovers from the previous iteration of the test that did not use the teapot dataset.

tests/integration/test_z_create_empty.py
```diff
@@ -0,0 +1,408 @@
...
+        return create_empty(
+            mdio_template=PostStack3DVelocityTemplate(data_domain="time", is_metric=True),
+            dimensions=dims,
+            output_path=output_path,
+            headers=headers,
+            overwrite=overwrite,
+        )
+
+    @classmethod
+    def validate_teapod_dataset_metadata(cls, ds: xr_Dataset, is_velocity: bool) -> None:
```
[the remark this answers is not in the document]
dmitriyrepin: Corrected. No "teapod" string(s) are present on the codebase anymore.

Issue #612, opened 2025-08-21
Grid Overrides


BrianMichell:
# AutoChannelWrap

We believe that this override does not require modification. Unit tests pass.

# NonBinned

### Input Parameters
- List of non-binned dimensions
- Chunk size of new dimension

### Behavior
From the template, we will:
1. Drop all dimensions contained in the list of non-binned dimensions
2. Add a new dimension `trace` to the N-1 position
    - The size of this dimension will be the product of all unique elements in the list of non-binned dimensions
3. Add one new coordinate per element in the list of non-binned dimensions        
   - These coordinates will have the dimensions up to `trace`

### Example
Template with `[shot_point, cable, channel, azimuth, offset, sample]` and non-binned `[azimuth, offset]`.

The new template will have dimensions `[shot_point, cable, channel, trace, sample]` and **new** coordinates `azimuth` and `offset` with dimensions `[shot_point, cable, channel, trace]` 

The chunk size of `sample` will be equal to the input parameter.

# DuplicateIndex

### Input Parameters
Just a flag (TBD).

### Behavior
We will add a new dimension called `trace` to the provided template at the N-1 position. This dimension **must** have a chunk size of 1. Coordinates will **not** be modified.

# ChannelWrap
We believe this can be safely removed as it is superseded by `AutoChannelWrap`.

- [x] Confirmation by @tasansal

# CalculateCable
We believe this can be safely removed as it is superseded by `AutoChannelWrap`.

- [x] Confirmation by @tasansal 

# AutoShotwrap
~~We believe this can be safely removed as it is not necessary.~~
This is necessary, but only for a specific schema.

- [x] Confirmation by @markspec

dmitriyrepin:
@tasansal , @BrianMichell 
For 'NonBinned':
If the original template had the following 
* dimensions ["shot_point", "cable", "channel", "time"]
* coordinate that is dependent on some those dimensions. e.g., "source_coord_x" ->  ["shot_point", "cable", "channel"]
what should happen to the coordinate (e.g., "source_coord_x") if declare the dimension "channel" NonBinned and remove the "channel" coordinate?

tasansal:
- `AutoChannelWrap` is merged via #632 
- `NonBinned` and `DuplicateIndex` were attempted via #633 however the implementation is wrong, needs to be redone.
- `CalculateCable` and `ChannelWrap` will be removed per reason @BrianMichell defined
[...]

PR #697, opened 2025-10-01
Remove CalculateCable grid override

BrianMichell:
Per #612 `CalculateCable` is no longer required. This PR removes associated source code, tests, and documentation regarding it.


tasansal (approved):
LGTM

PR #700, opened 2025-10-03
Ensure header overrides functionality of `TGSAI/segy` is available and applied.

tasansal:
# Summary
This PR enables the use of SEG-Y header overrides during the conversion from SEG-Y to MDIO format. Now, it is passed to the SegyFile constructor, leveraging the functionality from the segy library.

This ensures that users can specify custom header overrides (e.g., for trace headers or other metadata) when ingesting SEG-Y files, improving flexibility for handling non-standard or corrupted files.

There were two major bugs:
1. coordinate scalar wasn't applied at all
2. horizontal units were incorrectly extracted and wrongly saved (e.g. never)

Fixes both.


src/mdio/segy/scalar.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def validate_segy_scalar(scalar: int) -> bool:
+        """Validate if coord scalar matches the seg-y standard."""
+        if not INT16_MIN <= scalar <= INT16_MAX:  # 16-bit signed int range
+            logger.debug("Coordinate scalar is larger than int16")
+            return False
+        if scalar == 0:
+            logger.debug("Coordinate scalar is zero")
+            return False
+        return abs(scalar) in VALID_COORD_SCALAR  # valid values
```
BrianMichell, line 53:
```suggestion
    def validate_segy_scalar(scalar: int) -> bool:
        """Validate if coord scalar matches the seg-y standard."""
        if abs(scalar) not in VALID_COORD_SCALAR:
            logger.debug("Expected %i to be in ± %s but was not", scalar, sorted(VALID_COORD_SCALAR))
            return False
        return True
```

src/mdio/converters/segy.py
```diff
@@ -291,16 +303,17 @@ def populate_non_dim_coordinates(
...
-def _get_horizontal_coordinate_unit(segy_headers: list[Dimension]) -> LengthUnitModel | None:
+def _get_horizontal_coordinate_unit(segy_info: SegyFileHeaderDump) -> LengthUnitModel | None:
     """Get the coordinate unit from the SEG-Y headers."""
-    name = TraceHeaderFieldsRev0.COORDINATE_UNIT.name.upper()
-    unit_hdr = next((c for c in segy_headers if c.name.upper() == name), None)
-    if unit_hdr is None or len(unit_hdr.coords) == 0:
+    measurement_system_code = segy_info.binary_header_dict[MEASUREMENT_SYSTEM_KEY]
+
+    if int(measurement_system_code) not in (1, 2):
+        logger.warning("Measurement system header is empty or corrupt. Can't extract coordinate unit.")
```
BrianMichell, line 311: Scary message!
```suggestion
        logger.warning("Unexpected value found for coordinate unit. Please double check your segy_spec for correct %s byte location.", MEASUREMENT_SYSTEM_KEY )
```

Maybe this should suggest `segy_spec` and `segy_header_overrides`?

src/mdio/segy/scalar.py
tasansal: done

PR #705, opened 2025-10-08
Implement validation for all required fields

BrianMichell:
Implements a very simple validator.

@tasansal Let me know if you want extensive testing.


tasansal (changes requested):
Thanks Brian, the logic looks good. However, the implementation can be simplified with by using `set`s. This will eliminate the looping etc. can you change it like that? thanks.

PR #722, opened 2025-10-17
Make templates more clear by removing spatial dim names, and also allow changing full chunk shape.

### LINKED ISSUE #626: Allow chunk_size run-time override with template schemas
Template schemas are useful standards for data types.  Why the layout is consistent across a lot of different data types often the use varies.  To align with potential downstream tasks it would useful if users could specify both a template schema and overrides for the chunksize.  This ensures standardization of Dimensions and Coordinates while allowing for performance for downstream applications.

tasansal:
Having spatial dim names separate from dim names was causing confusion for templates; we removed spatial dim names all together.

Closes #626

### DIFF
--- a/tests/unit/v1/templates/test_seismic_templates.py
+++ b/tests/unit/v1/templates/test_seismic_templates.py
@@ -12,19 +12,19 @@ class TestSeismicTemplates:
 
     def test_chunk_shape_assignment(self) -> None:
         """Test that chunk shape is assigned correctly."""
-        tpl = Seismic2DPostStackTemplate("time")
-        tpl.build_dataset("test", (50, 50))
-        tpl.full_chunk_shape = (32, 32)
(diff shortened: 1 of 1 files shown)


src/mdio/segy/parsers.py
```diff
@@ -26,7 +26,7 @@
 def parse_headers(
     segy_file_kwargs: SegyFileArguments,
     num_traces: int,
-    subset: list[str] | None = None,
+    subset: tuple[str, ...] | None = None,
```
BrianMichell: It looks like changing from list to tuple might have some downstream issues?
https://github.com/TGSAI/mdio-python/blob/2e6f8290d60f218f6349baa444d4dedcb566f0e3/src/mdio/segy/_workers.py#L60-L62

PR #723, opened 2025-10-17
Implement full-dimension chunking support in Templates

### LINKED ISSUE #626: Allow chunk_size run-time override with template schemas

BrianMichell:
Resolves #626 by applying appropriate logic.

Tested that code below sets the `amplitude` Variable chunk size to `(64, 64, 1501)` where 1501 is discovered at runtime.

```python
if __name__ == "__main__":
    from mdio import segy_to_mdio
    from segy.schema import HeaderField, ScalarType
    from segy.standards import get_segy_standard
    from mdio.builder.template_registry import get_template
    import os

    os.environ["MDIO__IMPORT__SAVE_SEGY_FILE_HEADER"] = "1"
    os.environ["MDIO__IMPORT__CPU_COUNT"] = "16"

    input_segy = "filt_mig.sgy"
    output_mdio = "filt_mig.mdio"

    input_spec = get_segy_standard(1)
    input_fields = [
        HeaderField(name="inline", byte=17, format=ScalarType.INT32),
        HeaderField(name="crossline", byte=13, format=ScalarType.INT32),
        HeaderField(name="cdp_x", byte=81, format=ScalarType.INT32),
        HeaderField(name="cdp_y", byte=85, format=ScalarType.INT32),
    ]
    input_spec = input_spec.customize(trace_header_fields=input_fields)

    template = get_template("PostStack3DTime")
    template.full_chunk_shape = (64, 64, -1)

    segy_to_mdio(
        segy_spec=input_spec,
        mdio_template=template,
        input_path=input_segy,
        output_path=output_mdio,
        overwrite=True,
    )
```

### DIFF
--- a/src/mdio/builder/templates/base.py
+++ b/src/mdio/builder/templates/base.py
@@ -148,14 +148,29 @@ class AbstractDatasetTemplate(ABC):
     @property
     def full_chunk_shape(self) -> tuple[int, ...]:
         """Returns the chunk shape for the variables."""
-        return copy.deepcopy(self._var_chunk_shape)
+        # If dimension sizes are not set yet, return the stored shape as-is
+        if len(self._dim_sizes) != len(self._dim_names):
+            return self._var_chunk_shape
+
+        # Expand -1 values to full dimension sizes
+        return tuple(
+            dim_size if chunk_size == -1 else chunk_size
+            for chunk_size, dim_size in zip(self._var_chunk_shape, self._dim_sizes, strict=False)
+        )
 
     @full_chunk_shape.setter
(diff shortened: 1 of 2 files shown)


tasansal:
@BrianMichell is this needed at all? What happens today on `main` if you specify chunks as `(..., -1)`. Upstream libraries typically can handle `-1` chunk size natively.

PR #794, opened 2026-03-23
Add crs and bin size metadata fields to VariableMetadata

keylgonz:
Adds three optional fields to VariableMetadata for seismic survey geometry:

- crs: Coordinate Reference System (e.g., "EPSG:32610")
- bin_inl: Inline bin size
- bin_xl: Crossline bin size

Includes test coverage.


src/mdio/builder/schemas/v1/variable.py
```diff
@@ -44,6 +44,21 @@ class VariableMetadata(CoordinateMetadata):
         description="Minimal summary statistics.",
     )
 
+    crs: str | None = Field(
```
BrianMichell, line 47: I don't think a string CRS will be sufficient here. We should have support for the most common expected systems. An example of making it more usable might be:
```json
{
  "CRS": {
    "type": "EPSG",
    "code": 1234
  }
}
```

There are probably several projection systems that we should support, if you could please take a look.

You can see how we might want to structure this as a set of enums similar to [unitsV1](https://mdio-python.readthedocs.io/en/stable/data_models/version_1.html#mdio.builder.schemas.v1.units.AngleUnitModel). We will want to leave an escape hatch for custom free-form CRS data as well.

src/mdio/builder/schemas/v1/variable.py
```diff
@@ -44,6 +44,21 @@ class VariableMetadata(CoordinateMetadata):
...
+        default=None,
+        description="Coordinate Reference System as an EPSG code (e.g., 'EPSG:4326').",
+    )
+
+    bin_inl: float | None = Field(
```
BrianMichell, line 52: `bin_inl` and `bin_xl` are too specific to a use-cases with 3D poststack data. I would recommend looking into adding it in [dimensions.py](https://github.com/TGSAI/mdio-python/blob/main/src/mdio/builder/schemas/dimension.py) and updating its key to be `nominal_bin` to support non-binned data as well.

BrianMichell (changes requested):

Issue #824, opened 2026-06-08
Template `ReceiverGathers3D` cannot calculate `shot_index`


BrianMichell:
This template should not be calculating the dimension. Below is a minimal fix for this template.

```python
class CustomReceiverGathers3D(Seismic3DReceiverGathersTemplate):
    """Custom 3D receiver gathers template without calculated shot index."""

    def __init__(self) -> None:
        super().__init__()

        self._dim_names = ("receiver", "shot_line", "shot_point", "time")
        self._calculated_dims = ()
        self._logical_coord_names = ()

    def _add_coordinates(self) -> None:
        # Wrap builder to intercept coordinate additions and map shot_index to shot_point
        original_builder = self._builder

        class BuilderProxy:
            def __init__(self, builder):
                self._builder = builder

            def __getattr__(self, name):
                return getattr(self._builder, name)

            def add_coordinate(self, name, *args, **kwargs):
                # dimensions is a keyword-only argument in MDIODatasetBuilder.add_coordinate
                dimensions = kwargs.get("dimensions")
                if dimensions is None and len(args) > 0:
                    # If passed as positional argument (though it's keyword-only in builder, 
                    # let's be safe and check both)
                    dimensions = args[0]
                    args = list(args)
                    if name == "shot_point":
                        args[0] = ("shot_point",)
                    else:
                        args[0] = tuple(
                            "shot_point" if dim == "shot_index" else dim
                            for dim in args[0]
                        )
                    args = tuple(args)
                elif dimensions is not None:
                    if name == "shot_point":
                        kwargs["dimensions"] = ("shot_point",)
                    else:
                        kwargs["dimensions"] = tuple(
                            "shot_point" if dim == "shot_index" else dim
                            for dim in dimensions
                        )
                    return self._builder.add_coordinate(name, *args, **kwargs)

            try:
                self._builder = BuilderProxy(original_builder)
                super()._add_coordinates()
            finally:
                self._builder = original_builder
```

BrianMichell:
Resolved by #825

PR #825, opened 2026-06-10
Remove `shot_index` from `ReceiverGathers3D` template

BrianMichell:
This template was never supposed to have a calculated `shot_index`. As such, this was not ingestible.


BrianMichell:
Resolved #824

PR #851, opened 2026-07-29
Add Continuous Receiver Gathers template and grid override

BrianMichell:
Proposes implementation for CRG template and custom override to support epoch-based sensor chunking

EDIT: This PR has been reset to a draft so we can implement better defined templates and re-evaluate the requirements for multi-receiver gathers.


andgineer:
@BrianMichell 
This branch is cut from main before #852, so anything pinning refs/pull/851/head resolves zarr unbounded, picks up 3.3.0, and dies at import:

pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema
for <class 'zarr.codecs.blosc.BloscCname'>

Could you merge main into the branch? It needs the rebase before it merges anyway..

PR #861, opened 2026-08-05
Add single node CRG template

BrianMichell:
Implements the CRG template. Currently uses a placeholder name until a better one can be settled on.


andgineer:
Default chunk is (1, 512, 4096) — that splits our ~30k-sample CRG traces on time.

We're on (1, 512, 32768) so a whole trace stays one chunk. Can you flip the template default to that?

andgineer:
Hit the BloscCname / pydantic import crash on this tip with zarr>=3.3.0.

Looks like the zarr pin (or the #852 Blosc fix) made it onto the multi-node CRG PR but not this one. 
Can you carry it here too?

BrianMichell:
> Hit the BloscCname / pydantic import crash on this tip with zarr>=3.3.0.
> 
> Looks like the zarr pin (or the #852 Blosc fix) made it onto the multi-node CRG PR but not this one. Can you carry it here too?

Problem resolved offline.

This particular issue seems to have been an environment issue and is unrelated to the actual PR.

